PR #19, opened 2020-01-15
Fix spelling error in var doc string

georgh:
Just a small fix, a ` was missing.

### DIFF
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -316,10 +316,10 @@ Compute the sample variance of collection `itr`.
 The algorithm returns an estimator of the generative distribution's variance
 under the assumption that each entry of `itr` is an IID drawn from that generative
 distribution. For arrays, this computation is equivalent to calculating
-`sum((itr .- mean(itr)).^2) / (length(itr) - 1)).
+`sum((itr .- mean(itr)).^2) / (length(itr) - 1))`.
 If `corrected` is `true`, then the sum is scaled with `n-1`,
 whereas the sum is scaled with `n` if `corrected` is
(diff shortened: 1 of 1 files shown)


src/Statistics.jl
```diff
@@ -316,10 +316,10 @@ Compute the sample variance of collection `itr`.
 The algorithm returns an estimator of the generative distribution's variance
 under the assumption that each entry of `itr` is an IID drawn from that generative
 distribution. For arrays, this computation is equivalent to calculating
-`sum((itr .- mean(itr)).^2) / (length(itr) - 1)).
+`sum((itr .- mean(itr)).^2) / (length(itr) - 1))`.
 If `corrected` is `true`, then the sum is scaled with `n-1`,
 whereas the sum is scaled with `n` if `corrected` is
-`false` with `n` the number of elements in `itr`.
+`false`. `n` is the number of elements in `itr`.
```
nalimilan: Thanks! The first change is indeed correct, but I'd say the original formulation was better here:
```suggestion
`false` with `n` the number of elements in `itr`.
```

mschauer: Perhaps: “where n is the” ...

mschauer:
```suggestion
`false` where `n` is the number of elements in `itr`.
```

nalimilan: Sorry for the delay. I can't apply your suggestion since I don't have the permission to push to your branch. Can you do it? Thanks!

georgh: Sorry for the delay, didnt notice the review :/

nalimilan:
Thanks!

PR #21, opened 2020-02-01
Enable Travis CI

nalimilan:


.travis.yml
```diff
@@ -0,0 +1,44 @@
...
+  - 1.4
+  - nightly
+
+notifications:
+  email: false
+
+script:
+  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
+  # Needed to take precedence on Julia's version of Statistics
+  - julia -e 'using UUIDs; write("Project.toml", replace(read("Project.toml", String), r"uuid = .*?\n" =>"uuid = \"$(uuid4())\"\n"))'
```
mschauer: Info: The same devise is used in https://github.com/JuliaLang/Pkg.jl/blob/master/.travis.yml

nalimilan: Yes I took it from there.

.travis.yml
```diff
@@ -0,0 +1,60 @@
+language: julia
+
+os:
```
ararslan: We should be able to structure this as
```yaml
os:
  - linux
  - osx
  - windows
arch:
  - amd64
  - i386
  - arm64
jobs:
  exclude:
    - os: osx
      arch: i386
    - os: osx
      arch: arm64
    - os: windows
      arch: arm64
  include:
    - stage: "Documentation"
      ...
```

ararslan: This will give us x86-64, i686, and AArch64 builds for Linux, x86-64 for macOS, and x86-64 and i686 for Windows. This kind of build matrix structure didn't used to be possible but it is now.

nalimilan: OK, great. Though that doesn't seem to give i686 Windows: https://travis-ci.com/JuliaLang/Statistics.jl/builds/147088090

Would you mind updating your post at https://discourse.julialang.org/t/julia-is-now-available-on-travis-windows-builds/27187? I had just copied the information I could find there.

ararslan: Hm. There might still be something weird with Travis' config validator (since i386 isn't an official thing, it's a Julia-specific extension), or the Julia script for Travis might not correctly handle i386 instead of x86.

I'll be out for a couple of weeks but I'll try to remember to update the post when I get back.

Issue #22, opened 2020-02-03
Integer overflow in mean()


kagalenko-m-b:
```
julia> mean([1,1]*typemax(Int))
-1
julia> versioninfo()
Julia Version 1.3.1
Commit 2d5741174c (2019-12-30 21:36 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i3 CPU         540  @ 3.07GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, westmere)
Environment:
  JULIA_NUM_THREADS = 8
```

nalimilan:
Yes, that's expected, just like `typemax(Int)+typemax(Int)` doesn't give the correct result. See https://docs.julialang.org/en/v1/manual/faq/#faq-integer-arithmetic-1.

kagalenko-m-b:
That would have been expected of the function with an integer result. However, mean()
returns float and having silent integer overflow within is setting up a user for a nasty
unexpected bugs. Perhaps, you could reconsider closing that issue?

nalimilan:
Discussion at https://discourse.julialang.org/t/mean-of-integers-overfloats-bug-or-expected-behaviour/34767/3.

stevengj:
@nalimilan, since this function produces a floating-point result, I don't agree that it should overflow for integer inputs.

(This is analogous to the `hypot` function — as long as the output is representable, we shouldn't overflow on intermediate calculations.)

stevengj:
For example, an implementation like this:
```
function _mean(A::AbstractArray, ::Colon)
    isempty(A) && return sum(A) / length(A) # or maybe: throw(ArgumentError("mean requires non-empty array"))
    n = length(A)
    T = typeof(first(A) / n)
    return sum(x -> convert(T, x), A) / length(A)
end
```

stevengj:
Note that the `mean(A, dims=...)` variant already does the promotion before summing, so it avoids overflow already:
```jl
julia> mean([1,1]*typemax(Int), dims=1)
1-element Array{Float64,1}:
 9.223372036854776e18
```

stevengj:
@kagalenko-m-b, want to take a crack at a pull request?  Just need to use the above implementation for `_mean` and then add a couple of tests.

kagalenko-m-b:
[...]

PR #25, opened 2020-03-01
Solve the overflow in mean() on integers by promoting accumulator

kagalenko-m-b:
This PR corrects the issue #22, the bug in mean() on integer inputs.
```
julia> x = [1,1]*typemax(Int);

julia> mean(x)
-1.0

julia> mean(x,dims=1)[]
2.147483647e9
```

Implementation and tests are largely as [proposed](https://github.com/JuliaLang/Statistics.jl/issues/22#issuecomment-587119323) by @stevengj

### DIFF
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -163,7 +163,7 @@ julia> mean(A, dims=2)
  1.5
  3.5
 ```
-""" 
+"""
 mean(A::AbstractArray; dims=:) = _mean(A, dims)
 
 _mean(A::AbstractArray{T}, region) where {T} = mean!(Base.reducedim_init(t -> t/2, +, A, region), A)
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -173,7 +173,12 @@ function _mean(f, A::AbstractArray, dims=:)
         n = mapreduce(i -> size(A, i), *, unique(dims); init=1)
     end
     x1 = f(first(A)) / 1
-    return sum(x -> _mean_promote(x1, f(x)), A, dims=dims) / n
+    result = sum(x -> _mean_promote(x1, f(x)), A, dims=dims)
+    if dims === (:)
+        result /= n
+    else
+        result ./= n
+    end
 end
 
 function mean(r::AbstractRange{<:Real})
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -167,7 +167,16 @@ julia> mean(A, dims=2)
 mean(A::AbstractArray; dims=:) = _mean(A, dims)
 
 _mean(A::AbstractArray{T}, region) where {T} = mean!(Base.reducedim_init(t -> t/2, +, A, region), A)
-_mean(A::AbstractArray, ::Colon) = sum(A) / length(A)
+_prom(x::Missing, y) = (x,y)
+_prom(x, y::Missing) = (x,y)
+_prom(x::Missing, y::Missing) = (x,y)
+_prom(x,y)=promote(x,y)
+function _mean(A::AbstractArray, ::Colon)
+    isempty(A) && return one(eltype(A))*NaN
+    n = length(A)
+    x1 = first(A) / n
+    return sum(x -> first(_prom(x,x1)), A) / n
+end
 
 function mean(r::AbstractRange{<:Real})
     isempty(r) && return oftype((first(r) + last(r)) / 2, NaN)
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -167,12 +167,12 @@ _mean_promote(x::T, y::S) where {T,S} = convert(promote_type(T, S), y)
 
 function _mean(f, A::AbstractArray, dims=:)
     isempty(A) && return sum(f, A, dims=dims)/0
-    if dims == Colon()
+    if dims === (:)
         n = length(A)
     else
         n = mapreduce(i -> size(A, i), *, unique(dims); init=1)
     end
-    x1 = f(first(A)) / n
+    x1 = f(first(A)) / 1
     return sum(x -> _mean_promote(x1, f(x)), A, dims=dims) / n
 end
 
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -161,9 +161,7 @@ julia> mean(A, dims=2)
  3.5
 ```
 """
-#mean(A::AbstractArray; dims=:) = _mean(identity, A, dims)
-mean(A::AbstractArray; dims=:) = _mean(A, dims)
-_mean(A::AbstractArray{T}, region) where {T} = mean!(Base.reducedim_init(t -> t/2, +, A, region), A)
+mean(A::AbstractArray; dims=:) = _mean(identity, A, dims)
 
 _mean(A::AbstractArray, ::Colon) = _mean(identity, A)
 
@@ -172,7 +170,7 @@ _mean_promote(x::T, y::S) where {T,S} = convert(promote_type(T, S), y)
 function _mean(f, A::AbstractArray, dims=:)
     isempty(A) && return sum(f, A, dims=dims)/0
     if dims == Colon()
-        n =length(A)
+        n = length(A)
     else
         n = mapreduce(i -> size(A, i), *, unique(dims); init=1)
     end
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -175,9 +175,9 @@ function _mean(f, A::AbstractArray, dims=:)
     x1 = f(first(A)) / 1
     result = sum(x -> _mean_promote(x1, f(x)), A, dims=dims)
     if dims === (:)
-        result /= n
+        return result / n
     else
-        result ./= n
+        return result ./= n
     end
 end
 
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -163,19 +163,19 @@ julia> mean(A, dims=2)
  1.5
  3.5
 ```
-"""
+""" 
 mean(A::AbstractArray; dims=:) = _mean(A, dims)
 
 _mean(A::AbstractArray{T}, region) where {T} = mean!(Base.reducedim_init(t -> t/2, +, A, region), A)
-_prom(x::Missing, y) = (x,y)
-_prom(x, y::Missing) = (x,y)
-_prom(x::Missing, y::Missing) = (x,y)
-_prom(x,y)=promote(x,y)
 function _mean(A::AbstractArray, ::Colon)
-    isempty(A) && return one(eltype(A))*NaN
+    isempty(A) && return sum(A)/0
     n = length(A)
     x1 = first(A) / n
-    return sum(x -> first(_prom(x,x1)), A) / n
+    _prom(x::T, y::S) where {T,S} = begin
+        R = promote_type(T, S)
+        return convert(R, x)
+    end
+    return sum(x->_prom(x,x1), A) / n
 end
 
 function mean(r::AbstractRange{<:Real})
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -41,14 +41,11 @@ julia> mean(skipmissing([1, missing, 3]))
 2.0
 ```
 """
-_mean_promote(x::T, y::S) where {T,S} = convert(promote_type(T, S), x)
-
 function mean(itr)
     if isempty(itr)
         return mean(identity, itr)
     end
-    x1 = first(itr)/1
-    return mean(x -> _mean_promote(x, x1), itr)
+    return _mean_promote(itr)
 end
 
 """
@@ -176,11 +173,13 @@ mean(A::AbstractArray; dims=:) = _mean(A, dims)
 
 _mean(A::AbstractArray{T}, region) where {T} = mean!(Base.reducedim_init(t -> t/2, +, A, region), A)
 
-function _mean(A::AbstractArray, ::Colon)
+_mean(A::AbstractArray, ::Colon) = _mean_promote(A)
+function _mean_promote(A)
     isempty(A) && return sum(A)/0
-    n = length(A)
+    n = length(collect(A))
     x1 = first(A) / n
-    return sum(x -> _mean_promote(x, x1), A) / n
+    _prom(x::T, y::S) where {T,S} = convert(promote_type(T, S), x)
+    return sum(x -> _prom(x, x1), A) / n
 end
 
 function mean(r::AbstractRange{<:Real})
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -66,12 +66,12 @@ function mean(f, itr)
     end
     count = 1
     value, state = y
-    f_value = f(value)/1.0
+    f_value = f(value)/1
     total = Base.reduce_first(Base.add_sum, f_value)
     y = iterate(itr, state)
     while y !== nothing
         value, state = y
-        total += f(value)
+        total += _mean_promote(total, f(value))
         count += 1
         y = iterate(itr, state)
     end
@@ -101,7 +101,7 @@ julia> mean(√, [1 2 3; 4 5 6], dims=2)
  2.2285192400943226
 ```
 """
-mean(f, A::AbstractArray; dims=:) = _mean_promote(A, map_funct=f, dims=dims)
+mean(f, A::AbstractArray; dims=:) = _mean(f, A, dims)
 
 """
     mean!(r, v)
@@ -161,22 +161,23 @@ julia> mean(A, dims=2)
  3.5
 ```
 """
+#mean(A::AbstractArray; dims=:) = _mean(identity, A, dims)
 mean(A::AbstractArray; dims=:) = _mean(A, dims)
-# mean(A::AbstractArray; dims=:) = _mean_promote(A, dims=dims)
-
 _mean(A::AbstractArray{T}, region) where {T} = mean!(Base.reducedim_init(t -> t/2, +, A, region), A)
 
-_mean(A::AbstractArray, ::Colon) = _mean_promote(A)
-function _mean_promote(A::AbstractArray; map_funct=identity, dims=:)
-    isempty(A) && return sum(map_funct.(A))/0
+_mean(A::AbstractArray, ::Colon) = _mean(identity, A)
+
+_mean_promote(x::T, y::S) where {T,S} = convert(promote_type(T, S), y)
+
+function _mean(f, A::AbstractArray, dims=:)
+    isempty(A) && return sum(f, A, dims=dims)/0
     if dims == Colon()
         n =length(A)
     else
         n = mapreduce(i -> size(A, i), *, unique(dims); init=1)
     end
-    x1 = map_funct(first(A)) / n
-    _prom(x::T, y::S) where {T,S} = convert(promote_type(T, S), map_funct(x))
-    return sum(x -> _prom(x, x1), A, dims=dims) / n
+    x1 = f(first(A)) / n
+    return sum(x -> _mean_promote(x1, f(x)), A, dims=dims) / n
 end
 
 function mean(r::AbstractRange{<:Real})
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -171,11 +171,8 @@ function _mean(A::AbstractArray, ::Colon)
     isempty(A) && return sum(A)/0
     n = length(A)
     x1 = first(A) / n
-    _prom(x::T, y::S) where {T,S} = begin
-        R = promote_type(T, S)
-        return convert(R, x)
-    end
-    return sum(x->_prom(x,x1), A) / n
+    _prom(x::T, y::S) where {T,S} = convert(promote_type(T, S), x)
+    return sum(x -> _prom(x, x1), A) / n
 end
 
 function mean(r::AbstractRange{<:Real})
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -67,7 +67,7 @@ function mean(f, itr)
     count = 1
     value, state = y
     f_value = f(value)/1
-    total = Base.reduce_first(Base.add_sum, f_value)
+    total = Base.reduce_first(+, f_value)
     y = iterate(itr, state)
     while y !== nothing
         value, state = y
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -163,8 +163,6 @@ julia> mean(A, dims=2)
 """
 mean(A::AbstractArray; dims=:) = _mean(identity, A, dims)
 
-_mean(A::AbstractArray, ::Colon) = _mean(identity, A)
-
 _mean_promote(x::T, y::S) where {T,S} = convert(promote_type(T, S), y)
 
 function _mean(f, A::AbstractArray, dims=:)
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -41,7 +41,15 @@ julia> mean(skipmissing([1, missing, 3]))
 2.0
 ```
 """
-mean(itr) = mean(identity, itr)
+_mean_promote(x::T, y::S) where {T,S} = convert(promote_type(T, S), x)
+
+function mean(itr)
+    if isempty(itr)
+        return mean(identity, itr)
+    end
+    x1 = first(itr)/1
+    return mean(x -> _mean_promote(x, x1), itr)
+end
 
 """
     mean(f::Function, itr)
@@ -167,12 +175,12 @@ julia> mean(A, dims=2)
 mean(A::AbstractArray; dims=:) = _mean(A, dims)
 
 _mean(A::AbstractArray{T}, region) where {T} = mean!(Base.reducedim_init(t -> t/2, +, A, region), A)
+
 function _mean(A::AbstractArray, ::Colon)
     isempty(A) && return sum(A)/0
     n = length(A)
     x1 = first(A) / n
-    _prom(x::T, y::S) where {T,S} = convert(promote_type(T, S), x)
-    return sum(x -> _prom(x, x1), A) / n
+    return sum(x -> _mean_promote(x, x1), A) / n
 end
 
 function mean(r::AbstractRange{<:Real})
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -61,7 +61,7 @@ julia> mean([√1, √2, √3])
 function mean(f, itr)
     y = iterate(itr)
     if y === nothing
-        return Base.mapreduce_empty_iter(f, Base.add_sum, itr,
+        return Base.mapreduce_empty_iter(f, +, itr,
                                          Base.IteratorEltype(itr)) / 0
     end
     count = 1
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -131,9 +131,9 @@ end
         @test mean(x, dims=2) == [typemax(T)]'
     end
     # Check that mean of integers does not cause catastrophic loss of accuracy
-    let x = fill(typemax(Int), 10)
+    let x = fill(typemax(Int), 10),a=tuple(x...)
         @test (mean(x) == mean(x, dims=1)[] == mean(float, x)
-               ≈ float(typemax(Int)))  # avoid integer overflow (#22)
+               == mean(a) ≈ float(typemax(Int)))  # avoid integer overflow (#22)
     end
     let x = rand(10000)  # mean should use sum's accurate pairwise algorithm
         @test mean(x) == sum(x) / length(x)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -138,11 +138,11 @@ end
     let x = rand(10000)  # mean should use sum's accurate pairwise algorithm
         @test mean(x) == sum(x) / length(x)
     end
-    @test mean(Number[1, 1.5, 2+3im]) == 1.5+1im # mixed-type array
+    @test mean(Number[1, 1.5, 2+3im]) === 1.5+1im # mixed-type array
     @test isequal(mean(Float64[]), NaN)
     @test isequal(mean(Int[]), NaN)
-    @inferred(Statistics.mean(Int[]))
-    @inferred(Statistics.mean(Float32[]))
+    @inferred mean(Int[])
+    @inferred mean(Float32[])
     @test isequal(typeof(mean(Float32[])), typeof(mean(Float32[1])))
 end
 
@@ -724,7 +724,7 @@ end
     x = Any[1, 2, 4, 10]
     y = Any[1, 2, 4, 10//1]
     @test var(x) === 16.25
-    @test var(y) == 65//4
+    @test var(y) === 16.25
     @test std(x) === sqrt(16.25)
     @test quantile(x, 0.5)  === 3.0
     @test quantile(x, 1//2) === 3//1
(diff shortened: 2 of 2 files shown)


src/Statistics.jl
```diff
@@ -167,7 +167,16 @@ julia> mean(A, dims=2)
 mean(A::AbstractArray; dims=:) = _mean(A, dims)
 
 _mean(A::AbstractArray{T}, region) where {T} = mean!(Base.reducedim_init(t -> t/2, +, A, region), A)
-_mean(A::AbstractArray, ::Colon) = sum(A) / length(A)
+_prom(x::Missing, y) = (x,y)
```
nalimilan: It doesn't sound right that `Missing` needs a special case here. There may be other types which would need the same treatment. Better do something like this:
```julia
 _prom(x, y) = convert(promote_type(typeof(x), typeof(y)), x)
```
(and don't call `first` below)

src/Statistics.jl
```diff
@@ -167,7 +167,16 @@ julia> mean(A, dims=2)
...
+_prom(x, y::Missing) = (x,y)
+_prom(x::Missing, y::Missing) = (x,y)
+_prom(x,y)=promote(x,y)
+function _mean(A::AbstractArray, ::Colon)
+    isempty(A) && return one(eltype(A))*NaN
```
nalimilan: Something more clever than this is needed to ensure type stability. For example with `mean(Float32[])` vs `mean(Float32[1])`. Maybe `Base.add_sum(zero(eltype(A)), zero(eltype(A))))/0`.

src/Statistics.jl
kagalenko-m-b: Good suggestion, shouldn't be trying to extend the implementation of Base functions.

src/Statistics.jl
```diff
@@ -163,11 +163,20 @@ julia> mean(A, dims=2)
...
+""" 
 mean(A::AbstractArray; dims=:) = _mean(A, dims)
 
 _mean(A::AbstractArray{T}, region) where {T} = mean!(Base.reducedim_init(t -> t/2, +, A, region), A)
-_mean(A::AbstractArray, ::Colon) = sum(A) / length(A)
+function _mean(A::AbstractArray, ::Colon)
+    isempty(A) && return sum(A)/0
+    n = length(A)
+    x1 = first(A) / n
+    _prom(x::T, y::S) where {T,S} = begin
```
stevengj: you can say `function _prom` ... for a multi-line inner function.

But maybe just:
```jl
let T = typeof(first(A)/n)
    return sum(x -> convert(promote_type(T, typeof(x)), x), A) / n
end

File: src/Statistics.jl [hunk 4]
```diff
@@ -163,11 +163,20 @@ julia> mean(A, dims=2)
  1.5
  3.5
 ```
-"""
+"""
```
nalimilan:
```suggestion
"""
```

test/runtests.jl
```diff
@@ -130,6 +130,20 @@ end
...
+               ≈ float(typemax(Int)))  # avoid integer overflow (#22)
+    end
+    let x = rand(10000)  # mean should use sum's accurate pairwise algorithm
+        @test mean(x) == sum(x) / length(x)
+    end
+    @test mean(Number[1, 1.5, 2+3im]) == 1.5+1im # mixed-type array
+    @test isequal(mean(Float64[]), NaN)
+    @test isequal(mean(Int[]), NaN)
+    @inferred(Statistics.mean(Int[]))
+    @inferred(Statistics.mean(Float32[]))
```
nalimilan:
```suggestion
    @inferred mean(Int[])
    @inferred mean(Float32[])
```

test/runtests.jl
```diff
@@ -130,6 +130,20 @@ end
...
     end
+    # Check that mean of integers does not cause catastrophic loss of accuracy
+    let x = fill(typemax(Int), 10)
+        @test (mean(x) == mean(x, dims=1)[] == mean(float, x)
+               ≈ float(typemax(Int)))  # avoid integer overflow (#22)
+    end
+    let x = rand(10000)  # mean should use sum's accurate pairwise algorithm
+        @test mean(x) == sum(x) / length(x)
+    end
+    @test mean(Number[1, 1.5, 2+3im]) == 1.5+1im # mixed-type array
```
nalimilan:
```suggestion
    @test mean(Number[1, 1.5, 2+3im]) === 1.5+1im # mixed-type array
```

test/runtests.jl
```diff
@@ -710,7 +724,7 @@ end
     x = Any[1, 2, 4, 10]
     y = Any[1, 2, 4, 10//1]
     @test var(x) === 16.25
-    @test var(y) === 65//4
+    @test var(y) == 65//4
```
nalimilan:
```suggestion
    @test var(y) === 16.25
```
Technically this qualifies as a breaking change I guess.

test/runtests.jl
```diff
@@ -130,6 +130,20 @@ end
         @test mean(identity, x) == mean(identity, g) == typemax(T)
         @test mean(x, dims=2) == [typemax(T)]'
     end
+    # Check that mean of integers does not cause catastrophic loss of accuracy
+    let x = fill(typemax(Int), 10)
+        @test (mean(x) == mean(x, dims=1)[] == mean(float, x)
+               ≈ float(typemax(Int)))  # avoid integer overflow (#22)
```
nalimilan:
```suggestion
        @test mean(x) == mean(x, dims=1)[] == mean(float, x)
             ≈ float(typemax(Int))  # avoid integer overflow (#22)
```

src/Statistics.jl
kagalenko-m-b: That breaks type inference for some reason

```
julia> function my_mean(A::AbstractArray)
           isempty(A) && return sum(A)/0
           n = length(A)
           x1 = first(A) / n
           let T = typeof(first(A)/n)
             return sum(x -> convert(promote_type(T, typeof(x)), x), A) / n 
           end
       end
my_mean (generic function with 2 methods)

julia> @inferred(my_mean(Float32[],:))
ERROR: return type Float32 does not match inferred return type Any
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] top-level scope at REPL[13]:1
```

test/runtests.jl
kagalenko-m-b: True, however it is not obvious to me whether one or the other is more "right"

test/runtests.jl
kagalenko-m-b: Wouldn't that become two statements without brackets, as Julia does not have line continuation syntax?

nalimilan: Ah, got it. Usually I put the operator at the end of the first line to avoid this problem.

test/runtests.jl
nalimilan: Yes, the point is that I'm not sure whether our policy allows changing things like this in a minor release.

kagalenko-m-b: Note that this change restores consistency
```
julia> var(y,dims=1)[];var(y,dims=1)[]===16.25
true
```
So maybe calling it "breaking" is too strong.

src/Statistics.jl
kagalenko-m-b: Seems to me either Julia type inference has an issue, or simply isn't smart enough in this case.

test/runtests.jl
```diff
@@ -130,6 +130,20 @@ end
...
+    end
+    let x = rand(10000)  # mean should use sum's accurate pairwise algorithm
+        @test mean(x) == sum(x) / length(x)
+    end
+    @test mean(Number[1, 1.5, 2+3im]) === 1.5+1im # mixed-type array
+    @test isequal(mean(Float64[]), NaN)
+    @test isequal(mean(Int[]), NaN)
+    @inferred mean(Int[])
+    @inferred mean(Float32[])
+    @test isequal(typeof(mean(Float32[])), typeof(mean(Float32[1])))
```
nalimilan:
```suggestion
    @test @inferred mean(Int[]) === 0.0/0
    @test @inferred mean(Float64[]) === 0.0/0
    @test @inferred mean(Float32[]) === 0.0f0/0
```

kagalenko-m-b:
Maybe the calls should be refactored so that they ultimately call  ``` _mean(A::AbstractArray, ::Colon)``?

Otherwise we will be duplicating the code which is a bad style and additional room for for error

src/Statistics.jl
```diff
@@ -41,7 +41,12 @@ julia> mean(skipmissing([1, missing, 3]))
 2.0
 ```
 """
-mean(itr) = mean(identity, itr)
+function mean(itr)
```
nalimilan: This change isn't OK:
- `collect(A)` called by `mean_promote` will make a copy, which should be avoided at all costs. So we really need two different methods, one for arrays, and one for general iterators.
- `mean(f, itr)` should do the same promotion as `mean(itr)`. There's no reason it should have a different behavior, and the equivalence between `mean(itr)` and `mean(identity, itr)` seems essential.

kagalenko-m-b: > This change isn't OK:
>     * `collect(A)` called by `mean_promote` will make a copy, which should be avoided at all costs. So we really need two different methods, one for arrays, and one for general iterators.

Isn't there a method to count the elements without making a copy? Having duplicate implementations is bad style and multiplies the possibilities for bugs, when those two implementations go out of sync.

>     * `mean(f, itr)` should do the same promotion as `mean(itr)`. 

That is something I have to question. ```mean(f, itr)``` is user **explicitly** specifying the promotion. Library shouldn't be second-guessing him in that case.

>  no reason it should have a different behavior, and the equivalence between `mean(itr)` and `mean(identity, itr)` seems essential

The second form means a user telling the function "don't do the accumulator promotion".

Having two similar

kagalenko-m-b: Right now, ```mean()``` on empty tuple throws MethodError and this behaviour is enshrined in tests. On the other hand, ```mean()``` on zero-length array returns NaN. That looks inconsistent to me.

nalimilan: > I added the call to `collect()` to count the elements of "skipmissing" arrays.
> Isn't there a method to count the elements without making a copy? Having duplicate implementations is bad style and multiplies the possibilities for bugs, when those two implementations go out of sync.

Unfortunately that's not possible, as some iterators do not allow going over their elements twice. So you need to count the elements as you process them.

> That is something I have to question. `mean(f, itr)` is user **explicitly** specifying the promotion. Should the function be second-guessing him in that case?
> 
> Anyways, that objection is easy to accommodate by adding an optional argument to `_mean_promote()` with the default value `identity`.

Well the user isn't really specifying the promotion, s/he's merely saying that elements should be transformed first. For example, it's common to call `mean(abs, x)`, and it could be surprising that the result overflows while `mean(x)` doesn't.

> Right now, `mean()` on empty tuple throws MethodError and this behaviour is enshrined in tests. On the other hand, `mean()` on zero-length array returns NaN. That looks inconsistent to me.

That's because an empty array provides the element type so you know what's the type of zero that you should use. But an empty tuple doesn't have any element type, so the only choice we have is throw an error. Fixing the inconsistency by also throwing an error for arrays would be really annoying.

kagalenko-m-b: That explanation makes sense, thanks.

kagalenko-m-b: I have implemented promotion for generic iterators and unified promotion for AbstractArray in one function. If you take a look at the lines #164-165, the commented out line 165 passes all tests for ```mean()```, but breaks some tests of ```var()```.

src/Statistics.jl
```diff
@@ -165,9 +162,22 @@ julia> mean(A, dims=2)
 ```
 """
 mean(A::AbstractArray; dims=:) = _mean(A, dims)
+# mean(A::AbstractArray; dims=:) = _mean_promote(A, dims=dims)
 
 _mean(A::AbstractArray{T}, region) where {T} = mean!(Base.reducedim_init(t -> t/2, +, A, region), A)
-_mean(A::AbstractArray, ::Colon) = sum(A) / length(A)
+
+_mean(A::AbstractArray, ::Colon) = _mean_promote(A)
+function _mean_promote(A::AbstractArray; map_funct=identity, dims=:)
```
nalimilan: Why not just make this be `_mean(f, A::AbstractArray, dims)` as before? All `mean` methods now promote, so it would be weird to have only one of them use that term.

src/Statistics.jl
```diff
@@ -165,9 +162,22 @@ julia> mean(A, dims=2)
...
+    isempty(A) && return sum(map_funct.(A))/0
```
nalimilan:
```suggestion
    isempty(A) && return sum(map_funct, A)/0
```
Also this doesn't work when `dims !== :`. So better have a separate method for `dims::Colon`. You can define `_prom` in the global scope to use it in all methods (but call it e.g. `_promote_mean`).

src/Statistics.jl
```diff
@@ -66,7 +66,7 @@ function mean(f, itr)
     end
     count = 1
     value, state = y
-    f_value = f(value)
+    f_value = f(value)/1.0
```
nalimilan: Shouldn't this be `1`?

Below,  `f(value)` should be replaced with `_prom(total, f(value))` to make sure promotion happens (in theory `+` could choose any type).

src/Statistics.jl
kagalenko-m-b: Wouldn't adding dims=... argument to sum on line 171 fix that for all cases?

src/Statistics.jl
```diff
@@ -164,10 +161,24 @@ julia> mean(A, dims=2)
  3.5
 ```
 """
+#mean(A::AbstractArray; dims=:) = _mean(identity, A, dims)
```
nalimilan:
```suggestion

```

src/Statistics.jl
```diff
@@ -164,10 +161,24 @@ julia> mean(A, dims=2)
...
 mean(A::AbstractArray; dims=:) = _mean(A, dims)
-
 _mean(A::AbstractArray{T}, region) where {T} = mean!(Base.reducedim_init(t -> t/2, +, A, region), A)
-_mean(A::AbstractArray, ::Colon) = sum(A) / length(A)
+
+_mean(A::AbstractArray, ::Colon) = _mean(identity, A)
+
+_mean_promote(x::T, y::S) where {T,S} = convert(promote_type(T, S), y)
+
+function _mean(f, A::AbstractArray, dims=:)
```
nalimilan: [As noted by @stevengj](https://github.com/JuliaLang/Statistics.jl/issues/22#issuecomment-587122377), promotion isn't necessary when `dims !== :`. So just keep the existing `_mean` method from above.

src/Statistics.jl
nalimilan: [As noted by @stevengj](https://github.com/JuliaLang/Statistics.jl/issues/22#issuecomment-587122377), promotion isn't necessary when `dims !== :`. So just keep the existing `_mean` method from above and restrict this one to `dims::Colon`.

nalimilan:
Can you add tests for `mean(f, A, dims=...)` methods?

src/Statistics.jl
kagalenko-m-b: This line replaces the next two. When uncommented, all tests for ```mean()``` pass, but many tests for ```var()``` break. Is it feasible to eliminate the lines #165,166 ?

src/Statistics.jl
kagalenko-m-b: > [As noted by @stevengj](https://github.com/JuliaLang/Statistics.jl/issues/22#issuecomment-587122377), promotion isn't necessary when `dims !== :`. 

Is that promotion exactly identical in all cases to _mean_promote()?

> Can you add tests for `mean(f, A, dims=...)` methods?

Right now there're two implementations for dims not equal to Colon - line #166 and the one starting at #172. They don't seem to be fully compatible. The new one breaks tests for ```var()```.

kagalenko-m-b: What I'm driving at, perhaps, the implementation from #166 ought to be eliminated. But that might require fixing errors in var().

nalimilan: > Is that promotion exactly identical in all cases to `_mean_promote()`?

Well the situation there is quite messy, and there's a very old comment saying that things should be improved. It calls `Base.reducedim_init`, which fills the array with `z = zero(zero(eltype(A))/2)); z+z` for common numeric types. For other types, it fills the array with `z = zero(zero(_realtype(f, promote_union(eltype(A))))/2); z+z` if `zero` is supported, and otherwise with `zero(sum(t -> T/2, A))`. Overall since `/2` is present in all paths, the promotion behavior should be relatively consistent with `_mean_promote`.

Things will probably differ for heterogeneous types, where `_mean_promote` will progressively promote the type to take into account new entries, while when `dims !== :` promotion will always be done upfront. But we making reduction over dimensions promote progressively would be complex and inefficient (that would imply copying arrays), and in practice cases where results would differ should be quite rare. OTOH we could make the `dims === :` case always promote beforehand using similar rules to `Base.reducedim_init`, but that's probably not worth it.

@stevengj Do you agree?

> Right now there're two implementations for dims not equal to Colon - line #166 and the one starting at #172. They don't seem to be fully compatible. The new one breaks tests for `var()`.

Can you show the failures you get? Indeed it would be nice if we were able to use a single implementation.

kagalenko-m-b: >> Is that promotion exactly identical in all cases to _mean_promote()?

> Well the situation there is quite messy, and there's a very old comment saying that things should be improved. 

Right now seems like a good opportunity to do so.

> Can you show the failures you get? 

You know, funny thing happened, I tried to reproduce the failures and all tests passed without complaint. I pushed the implementation not reliant on  ```Base.reducedim_init()```, take a look.

src/Statistics.jl
```diff
@@ -164,10 +161,20 @@ julia> mean(A, dims=2)
...
-mean(A::AbstractArray; dims=:) = _mean(A, dims)
+mean(A::AbstractArray; dims=:) = _mean(identity, A, dims)
+
+_mean_promote(x::T, y::S) where {T,S} = convert(promote_type(T, S), y)
 
-_mean(A::AbstractArray{T}, region) where {T} = mean!(Base.reducedim_init(t -> t/2, +, A, region), A)
-_mean(A::AbstractArray, ::Colon) = sum(A) / length(A)
+function _mean(f, A::AbstractArray, dims=:)
+    isempty(A) && return sum(f, A, dims=dims)/0
+    if dims == Colon()
```
nalimilan, line 170: This should help the compiler to simplify code based on types:
```suggestion
    if dims === :
```

src/Statistics.jl
```diff
@@ -164,10 +161,20 @@ julia> mean(A, dims=2)
...
+        n = length(A)
+    else
+        n = mapreduce(i -> size(A, i), *, unique(dims); init=1)
+    end
+    x1 = f(first(A)) / n
```
nalimilan, line 175: Since the exact value of `n` doesn't matter here, better use `1` as in the method above:
```suggestion
    x1 = f(first(A)) / 1
```

test/runtests.jl
```diff
@@ -130,6 +130,18 @@ end
         @test mean(identity, x) == mean(identity, g) == typemax(T)
         @test mean(x, dims=2) == [typemax(T)]'
     end
+    # Check that mean of integers does not cause catastrophic loss of accuracy
```
nalimilan, line 133: Can you add similar tests for the method operating on general iterables? You can just use `sum(v for v in x)`.

kagalenko-m-b:
[Script](https://pastebin.com/hGHg63qz) to benchmark ```mean()``` gives those [results](https://pastebin.com/jRJ1d351)

nalimilan:
Thanks. So there are no regressions for `Float64` (as it should). For `Int64` on 64-bit, the PR is about 50% slower, which sounds acceptable. For `Int32` on 32-bit, it's about 2× to 3× slower. I guess the penalty will be much higher for e.g. `Int8` due to SIMD, but that's inevitable.

Sorry to be a pain, but could you run benchmarks 1-4 on `(v for v in x)` and `(v for v in y)` to time the generic iterators method?

src/Statistics.jl
```diff
@@ -167,12 +167,12 @@ _mean_promote(x::T, y::S) where {T,S} = convert(promote_type(T, S), y)
 
 function _mean(f, A::AbstractArray, dims=:)
     isempty(A) && return sum(f, A, dims=dims)/0
-    if dims == Colon()
+    if dims === (:)
```
nalimilan, line 170: Parentheses are unnecessary.
```suggestion
    if dims === :
```

kagalenko-m-b: Without brackets compiler gives the error
> ERROR: LoadError: syntax: space not allowed after ":" used for quoting

nalimilan: Ah, sorry, I had checked `dims === 1` at the REPL and it works there. Not sure why not after `if`.

kagalenko-m-b:
Updated [benchmark](https://pastebin.com/4eXtL766) gives those [results](https://pastebin.com/LdPrxRJv)

StefanKarpinski:
What about accumulating to `Int64` for smaller integer types? That should avoid the overflow problem without losing precision and it might preserve the SIMD properties.

kagalenko-m-b:
> What about accumulating to `Int64` for smaller integer types?

That would be Int32 on 32 bit systems, presumably. Implementation-wise, does not seem to be too difficult, just change line 175
```
x1 = f(first(A)) / 1
```
to something more involved.

StefanKarpinski:
> That would be Int32 on 32 bit systems, presumably.

Even 32-bit systems have 64-bit integers, so no, I'd think you'd want to do it the same way everywhere. The main reason that `Int` is used on 32-bit systems is because those systems only have 32-bits of addressable memory so using 64-bit indices is a waste.

kagalenko-m-b:
Another issue with that is that still runs the risk of integer overflow if the array length is larger than 2^32, so the implementation presumably would need to check for that as well. And what about the iterators?

nalimilan:
Yes then summing you have no idea what's the upper bound. I guess `typemax(Int8)*typemax(Int)` is the practical limit, but how much below that maximum can we expect to be in general?

IMHO accumulating in a different type from the output only for some input types would makes the function's behavior quite hard to predict.

StefanKarpinski:
> Another issue with that is that still runs the risk of integer overflow if the array length is larger than 2^32, so the implementation presumably would need to check for that as well.

What about having two accumulation levels:

* accumulate up to 2^32 values as an Int64
* accumulate Int64 values using Float64

More generally, it might also be possible to use exact arithmetic until an overflow is detected and then switch to float.

kagalenko-m-b:
> What about having two accumulation levels:
> 
>     * accumulate up to 2^32 values as an Int64
> 
>     * accumulate Int64 values using Float64

On AbstractArrays this would be easy to implement, on iterators that would add an "if"  into every loop iteration.

> More generally, it might also be possible to use exact arithmetic until an overflow is detected and then switch to float.

AFAIK hardware detection (inexact or overflow exception with an attached handler) is possible in IEEE754 arithmetic, but not in integer, at least, not on every CPU. So the only way to implement this functionality would be to insert an "if" into every addition operation after the 2^32th.

nalimilan:
OK, now I see the idea. That would just be an optimization that wouldn't affect the results, so we could apply it only to `AbstractArray` and iterators for which `Base.IteratorSize(itr) isa Union{Base.HasLength, Base.HasShape} && Base.IteratorEltype(itr) isa Base.HasEltype`. Then you could simply call `sum` as an optimization if the following condition is fulfilled:
```julia
T = eltype(itr)
isconcrete(nonmissingtype(T)) && T <: Union{Real, Missing} &&
    typemax(T) * length(itr) < min(typemax(Base.add_sum(zero(T), zero(T))), maxintfloat(zero(T)/1) &&
    typemin(T) * length(itr) > max(typemin(Base.add_sum(zero(T), zero(T))), -maxintfloat(zero(T)/1)
```
Instead of `Union{Real, Missing}`, we could also use `Union{Base.SmallSigned, Base.SmallUnsigned}`, so that we are sure the optimization isn't used for custom types for which it may not be correct (not sure if that's plausible).

@StefanKarpinski Does that make sense?

StefanKarpinski:
That sounds better thought through that what I had come up with but yes, you can break a list into small enough pieces that each one can't overflow and then sum those recursively.

nalimilan:
Actually I hadn't realized you could break arrays in chunks. In practice, is it really necessary? Even for `Int32`, the condition I wrote in my previous comment works for up to 4e9 entries. We can leave splitting as a future optimization.

stevengj:
I'm skeptical that trying to optimize the small-integer case is worth the effort (= code complexity, fragility).   For what real application is `mean` the performance-limiting step, such that a constant-factor slowdown is a big deal?

kagalenko-m-b:
That is my feeling on this matter, as well.

nalimilan:
No idea. But adding that fast path to avoid a performance regression isn't very costly either. Also, you never know what kind of operation people are going to benchmark against other languages.

stevengj:
> But adding that fast path to avoid a performance regression isn't very costly either.

It is if you get it wrong.

nalimilan:
@stevengj Are you OK with the PR as it is then?

src/Statistics.jl
```diff
@@ -164,10 +161,20 @@ julia> mean(A, dims=2)
...
+function _mean(f, A::AbstractArray, dims=:)
+    isempty(A) && return sum(f, A, dims=dims)/0
+    if dims === (:)
+        n = length(A)
+    else
+        n = mapreduce(i -> size(A, i), *, unique(dims); init=1)
+    end
+    x1 = f(first(A)) / 1
+    return sum(x -> _mean_promote(x1, f(x)), A, dims=dims) / n
+end
```
stevengj, line 177: Maybe avoid creating a second array:
```jl
result = sum(x -> _mean_promote(x1, f(x)), A, dims=dims)
return (result .*= 1/n)
```
This won't work in the `dims === (:)` case, but in that case you could just return `mean(f, A)`.

kagalenko-m-b: Sorry, I don't follow - `mean(f,A)` is just `_mean(f, A, :)`

stevengj: Whoops, I guess  in the `dims === (:)` case return `sum(x -> _mean_promote(x1, f(x)), A) / length(A)`

test/runtests.jl
```diff
@@ -130,6 +130,19 @@ end
...
+    # Check that mean avoids integer overflow (#22)
+    let x = fill(typemax(Int), 10), a = tuple(x...)
+        @test (mean(x) == mean(x, dims=1)[] == mean(float, x)
+               == mean(a) == mean(v for v in x)  == mean(v for v in a)
+               ≈ float(typemax(Int)))
+    end
+    let x = rand(10000)  # mean should use sum's accurate pairwise algorithm
+        @test mean(x) == sum(x) / length(x)
+    end
+    @test mean(Number[1, 1.5, 2+3im]) === 1.5+1im # mixed-type array
```
nalimilan, line 142: While we're at it, ths test should be applied to `mean(v for v in [...])` too to check the generic iterator method.

Same for `@inferred` below. Maybe use `Iterators.filter(x -> true, Int[])` and so on since it has an eltype, so it also tests the empty case.

src/Statistics.jl
kagalenko-m-b: OK, I feel like I'm taking crazy pills here.
```
julia> x=[1//1];x .*= 1/5
1-element Array{Rational{Int64},1}:
 3602879701896397//18014398509481984

julia> versioninfo()
Julia Version 1.4.0
Commit b8e9a9ecc6 (2020-03-21 16:36 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i3 CPU         540  @ 3.07GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-8.0.1 (ORCJIT, westmere)
Environment:
  JULIA_NUM_THREADS = 8
```

But on 32 bit system
```
julia> x=[1//1];x .*= 1/5
1-element Array{Rational{Int32},1}:
 1//5

julia> versioninfo()
Julia Version 1.4.0
Commit b8e9a9ecc6 (2020-03-21 16:36 UTC)
Platform Info:
  OS: Linux (i686-pc-linux-gnu)
  CPU: Intel(R) Pentium(R) 4 CPU 3.00GHz
  WORD_SIZE: 32
  LIBM: libopenlibm
  LLVM: libLLVM-8.0.1 (ORCJIT, prescott)
```

kagalenko-m-b: [Tests](https://travis-ci.com/github/kagalenko-m-b/RationalDivision/builds/159956125)of this bug  on different architectures and Julia versions

nalimilan: That's just due to rounding I guess: `3602879701896397//18014398509481984 ≈ 1/5`. Try with `x .*= 1//5` if you want the exact result.

test/runtests.jl
```diff
@@ -140,9 +140,13 @@ end
...
-    @test (@inferred mean(Int[])) === 0.0/0
-    @test (@inferred mean(Float64[])) === 0.0/0
-    @test (@inferred mean(Float32[])) === 0.0f0/0
+    @test mean(v for v in Number[1, 1.5, 2+3im]) === 1.5+1im
+    @test (@inferred mean(Int[])) === 0/0
+    @test (@inferred mean(Float32[])) === 0.f0/0    
+    @test (@inferred mean(Float64[])) === 0/0
+    @test (@inferred mean(filter(x -> true, Int[]))) === 0/0
+    @test (@inferred mean(filter(x -> true, Float32[]))) === 0.f0/0
+    @test (@inferred mean(filter(x -> true, Float64[]))) === 0/0
```
nalimilan, line 149: `filter` returns a vector, you need `Iterators.filter`. Probably also `using Iterators` at the top of the file.
```suggestion
    @test (@inferred mean(Iterators.filter(x -> true, Int[]))) === 0/0
    @test (@inferred mean(Iterators.filter(x -> true, Float32[]))) === 0.f0/0
    @test (@inferred mean(Iterators.filter(x -> true, Float64[]))) === 0/0
```

src/Statistics.jl
```diff
@@ -173,7 +173,12 @@ function _mean(f, A::AbstractArray, dims=:)
         n = mapreduce(i -> size(A, i), *, unique(dims); init=1)
     end
     x1 = f(first(A)) / 1
-    return sum(x -> _mean_promote(x1, f(x)), A, dims=dims) / n
+    result = sum(x -> _mean_promote(x1, f(x)), A, dims=dims)
+    if dims === (:)
+        result /= n
+    else
+        result ./= n
+    end
```
nalimilan, line 181:
```suggestion
    if dims === (:)
        return result / n
    else
        return result ./= n
    end
```

src/Statistics.jl
kagalenko-m-b: Yeah, I get it now:
```
julia> Rational{Int64}[0.2]
1-element Array{Rational{Int64},1}:
 3602879701896397//18014398509481984

julia> Rational{Int32}[0.2]
1-element Array{Rational{Int32},1}:
 1//5

```

nalimilan:
@StefanKarpinski Your call?

StefanKarpinski:
I think this is not my call, but if you and @stevengj are happy with this, I say go ahead.

nalimilan:
OK. Merging then. Adding fast paths for arrays of small integers can be discussed later.

@kagalenko-m-b Thanks for finishing this!

kagalenko-m-b:
Took longer than I expected, but in the end we much improved the original version.

PR #30, opened 2020-03-13
Make covariance and correlation work for any iterators

pdeffebach:
This PR is in reference to #35050 in the Julia Repo [here](https://github.com/JuliaLang/julia/issues/35050). The goal is to make working with iterators more convenient. In particular, this PR will make working with iterators that skip over missing values more convenient. 

This PR will improve both `cov` and `cor`. However it starts with `cov`, `covm` and `covzm`. I think I went overboard in avoiding allocations because `covm` and `covzm` are handeled via `iterate` only, the way `mean(itr)` is handled in the same file. However `covm` for `AbstractArrays` is fine to use `map` and allocate a new vector. The decision on how to proceed will be based on how we would like to handle stateful iterators. 

Currently there are 4 new methods added. 

* `covzm(itr::Any; corrected::Bool=true)`
* `covzm(x::Any, y::Any, corrected::Bool=true)`
* `covm(itr::Any, itrmean; corrected::Bool=true)`
* `covm(x::Any, xmean, y::Any, ymean; corrected::Bool=true)`

As a consequence, something like 

```
cov(x::AbstractMatrix, itr::Any)
```

will break. 

This is my first PR to a stdlib, so let me know if I am missing any conventions. Thanks!

### DIFF
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -487,7 +487,6 @@ _conjmul(x, y) = x * _conj(y)'
 
 # core functions
 
-unscaled_covzm(x::AbstractVector{<:Number})    = sum(_abs2, x)
 unscaled_covzm(x::AbstractVector)              = sum(_abs2, x)
 unscaled_covzm(x::AbstractMatrix, vardim::Int) = (vardim == 1 ? _conj(x'x) : x * x')
 
@@ -503,13 +502,13 @@ unscaled_covzm(x::AbstractMatrix, y::AbstractMatrix, vardim::Int) =
 function covzm(itr::Any; corrected::Bool=true)
     y = iterate(itr)
     if y === nothing
-        return Base.mapreduce_empty_iter(_abs2, Base.add_sum, itr,
-                                         Base.IteratorEltype(itr)) / 0
+        v = _abs2(zero(eltype(itr)))
+        return (v + v) / 0
     end
     count = 1
     value, state = y
     f_value = _abs2(value)
-    total = Base.reduce_first(Base.add_sum, f_value)
+    total = Base.reduce_first(+, f_value)
     y = iterate(itr, state)
     while y !== nothing 
         value, state = y
@@ -532,23 +531,21 @@ function covzm(x::Any, y::Any; corrected::Bool=true)
     z = zip(x, y)
     z_itr = iterate(z)
     if z_itr === nothing
-        # TODO: Understand how to improve this error.
-        #return Base.mapreduce_empty_iter(t -> _conj(t[2])*t[1]', Base.add_sum, itr,
-        #                                 Base.IteratorEltype(x)) / 0
-        return NaN
+        v = _conjmul(zero(eltype(x)), zero(eltype(y)))
+        return (v + v) / 0
     end
     count = 1
-    value, state = z_itr
-    f_value = _conjmul(value[1], value[2])
-    total = Base.reduce_first(Base.add_sum, f_value)
+    (xi, yi), state = z_itr
+    f_value = _conjmul(xi, yi)
+    total = Base.reduce_first(+, f_value)
     z_itr = iterate(z, state)
     while z_itr !== nothing 
-        value, state = z_itr
-        total += _conjmul(value[1], value[2])
+        (xi, yi), state = z_itr
+        total += _conjmul(xi, yi)
         count += 1
         z_itr = iterate(z, state)
     end
-    return total ./ (count - Int(corrected))
+    return total / (count - Int(corrected))
 end
 covzm(x::AbstractVector, y::AbstractVector; corrected::Bool=true) =
     unscaled_covzm(x, y) / (length(x) - Int(corrected))
@@ -563,24 +560,21 @@ end
 
 # covm (with provided mean)
 ## Use map(t -> t - xmean, x) instead of x .- xmean to allow for Vector{Vector}
-## which can't be handled by broadcastz
+## which can't be handled by broadcast
 function covm(itr::Any, itrmean; corrected::Bool=true)
     y = iterate(itr)
-    f = let itrmean = itrmean
-        x -> _abs2(x-itrmean)
-    end
     if y === nothing
-        return Base.mapreduce_empty_iter(f, Base.add_sum, itr,
-                                         Base.IteratorEltype(itr)) / 0
+        v = _abs2(zero(eltype(itr - itrmean)))
+        return (v + v) / 0
     end
     count = 1
-    value, state = y
-    f_value = f(value)
-    total = Base.reduce_first(Base.add_sum, f_value)
+    itri, state = y
+    first_value = _abs2(itri - itrmean)
+    total = Base.reduce_first(+, first_value)
     y = iterate(itr, state)
     while y !== nothing 
-        value, state = y
-        total += f(value)
+        itri, state = y
+        total += _abs2(itri - itrmean)
         count += 1
         y = iterate(itr, state)
     end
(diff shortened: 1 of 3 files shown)


nalimilan:
Thanks. That implementation is reasonable, but I wonder whether using the existing methods based on `map` (just widening the signatures) wouldn't be better given how simple that would be. That will allocate more, but in terms of performance it might be faster due to the use of BLAS on the collected vectors. OTOH, I generally prefer avoiding allocations where possible so...

Could you run a few benchmarks since you've already written the code? For example, testing on generators and `skipmissing[s]` would be interesting.

> As a consequence, something like
> 
> ```
> cov(x::AbstractMatrix, itr::Any)
> ```
> 
> will break.

Why would it break?

pdeffebach:
> Why would it break?

It won't "break" but it won't give you the matrix variance-covariance matrix from treating each column in the matrix as a separate vector. 

It will fall back on `iterate`-ing the matrix and thus the `itr` would have to be an iterator of scalars with the same number of elements as the corresponding matrix. This might be unintuitive behavior.

pdeffebach:
One thing I wasn't expecting was that this fails:

```
x = [rand(5) for i in 1:10]
y = [rand(5) for i in 1:10]
cov(x, y)
# long error message
```
Whereas

* If `x` and `y` are matrices where each row is one of their vectors, `cov` works. 
* If `x` and `y` are generators, then after this PR it works.

pdeffebach:
Here are the results of my benchmark. I added a method so that the above covariance between vectors of vectors works. This makes benchmarking easier but will also have to be added since iterators and vectors should behave similarly. 

Highlights: 

* Things look good. Slowdown from vectors to generators less than a factor of 10. 
* covariances between `skipmissing`s is just as fast as `collect`ing them. But this is due to the overhead with iteration. 
* Not `collect`ing generators is faster. 
* Multivariate covariances, i.e. iterators of vectors, or two matrices, are quite bad. Slowdown above a factor of 10 compared to matrix case. The two calls actually have the same number of allocations, which I don't understand. Also, my implementation is wrong. I am returning the transpose of what I want. I need to look into this. 

Unfortunately given these difficulties this PR is unlikely to be merged before the 1.5 feature freeze. 


```
============
Proportion missing: 0.2
============

Covariance:
============
N = 10000: vectors
    7.8364000000e-05 seconds

N = 10000: skipmissings, no collect
    4.9105100000e-04 seconds

N = 10000: skipmissings, collect
    4.7475300000e-04 seconds

N = 10000: generator, no collect
    6.5641000000e-05 seconds

N = 10000: generator, collect
    1.2176000000e-04 seconds

N = 10000, multivariate vector, 2 args
    1.7621935000e-02 seconds

N = 10000, multivariate matrix, 2 args
    6.8707000000e-04 seconds

N = 10000, multivariate generator, no collect
    1.6903689000e-02 seconds

N = 10000, multivariate generator, collect
    1.7486476000e-02 seconds

N = 10000, multivariate vector, 1 arg
    1.5073024000e-02 seconds

N = 10000, multivariate matrix, 1 arg
    4.1739900000e-04 seconds

N = 10000, multivariate generator, 1 arg, no collect
    1.4867595000e-02 seconds

N = 10000, multivariate generator, 1 arg, collect
    1.4857078000e-02 seconds
```

src/Statistics.jl
```diff
@@ -517,17 +560,75 @@ end
 
 # covm (with provided mean)
 ## Use map(t -> t - xmean, x) instead of x .- xmean to allow for Vector{Vector}
-## which can't be handled by broadcast
+## which can't be handled by broadcastz
```
nalimilan, line 563:
```suggestion
## which can't be handled by broadcast
```
(but this comment just be moved below)

src/Statistics.jl
```diff
@@ -517,17 +560,75 @@ end
...
+function covm(x::Any, xmean, y::Any, ymean; corrected::Bool=true)
+    z = zip(x, y)
+    z_itr = iterate(z)
+    if z_itr === nothing
+        # TODO: Understand how to improve this error.
+        #return Base.mapreduce_empty_iter(t -> _conj(t[2])*t[1]', Base.add_sum, itr,
+        #                                 Base.IteratorEltype(x)) / 0
+        return NaN
+    end
+    f = let xmean = xmean, ymean = ymean
```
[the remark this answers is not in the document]
pdeffebach: I would have to write `(x, y) -> _conj(x - xmean, y - ymean)`. Wouldn't that cause the closure bug?

pdeffebach:
@nalimilan if you are happy with this I can move on to `cor` and `var` and any other functions that need it.

src/Statistics.jl
```diff
@@ -504,6 +525,28 @@ function covzm(x::AbstractMatrix, vardim::Int=1; corrected::Bool=true)
     A .= A .* b
     return A
 end
+function covzm(x::Any, y::Any; corrected::Bool=true)
```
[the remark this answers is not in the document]
nalimilan: Have you tried this?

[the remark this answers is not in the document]
pdeffebach: Just did. 

```
julia> x = randn(10_000); y = randn(10_000);

julia> x = x .- Statistics.mean(x); y = y .- Statistics.mean(y);

julia> gx = (xi for xi in x); gy = (yi for yi in y);

julia> @btime Statistics.covzm($gx, $gy)
  11.820 μs (0 allocations: 0 bytes)
0.009480946018135657

julia> @btime Statistics.covm($gx, 0,  $gy, 0)
  14.768 μs (0 allocations: 0 bytes)
0.009480946018135657
```

Let me know what you think of that difference. It doesn't seem very important to me. The `covzm` version takes 80% as long as `covm` with `0`s.

nalimilan:
> ```julia> x = randn(10_000); y = randn(10_000);
>
> julia> x = x .- Statistics.mean(x); y = y .- Statistics.mean(y);
>
> julia> gx = (xi for xi in x); gy = (yi for yi in y);
>
> julia> @btime Statistics.covzm($gx, $gy)
>  11.820 μs (0 allocations: 0 bytes)
> 0.009480946018135657
>
> julia> @btime Statistics.covm($gx, 0,  $gy, 0)
>  14.768 μs (0 allocations: 0 bytes)
> 0.009480946018135657
>```
> Let me know what you think of that difference. It doesn't seem very important to me. The covzm version takes 80% as long as covm with 0s.

I think it's OK. We could get rid of that small overhead by defining an internal helper with `@inline`, but I guess that's not worth it (and the compiler may improve in the future).

pdeffebach:
Working on correlation now. Shouldn't be too hard.

pdeffebach:
I am working through this. I want to add `corzm(x::Any, y::Any)`.

There seems to be no `corzm(x::AbstractVector, y::AbstractVector)`, so I won't add the method above, which is analogous. Is this intentional? Do we not care about the missing method because it is not exported?

Let me know if I should cc Andreas or someone else who has worked with this code more.

nalimilan:
I think that method isn't needed since this definition works both for vectors and matrices:
```julia
corm(x::AbstractVecOrMat, xmean, y::AbstractVecOrMat, ymean, vardim::Int=1) =
    corzm(x .- xmean, y .- ymean, vardim)
```

Looking at how tricky the changes need to be to support iterators, I'm more and more inclined to just add methods to `cor` which call `collect` on the iterators. Anyway the current code already makes a copy to subtract the mean, so if you subtract the mean in-place you won't be less efficient than the code for vectors. Calling `collect` has the advantage that you will know the element type, which isn't the case for non-`HasEltype` iterators: in that case, writing correct code could be really difficult and possibly ugly.

pdeffebach:
I am starting to agree. I was doing well using the `iterate` methods for everything, but ran into

1. Using `first(itr)` to do dispatch, which will have compilation issues
2. Bugs relating to `Any[1, 2]`, where my dispatch currently relies assuming an vector that isn't full of numbers is full of vectors. 
3. Not working with stateful iterators anyways because of `mean`, as you describe above. Though `corm` should still work with stateful iterators. 

Tests currently fail due to bug 2, and writing generic code means adding a lot of methods to helper functions. It probably isn't worth it. Take a look at the most recent commit pushed to this branch to make the final call based off of implementation that is pretty much what we want if we avoid `collect`ing, barring a few helper method dispatches.

nalimilan:
I vote for keeping things simple, at least for now. :-)

src/Statistics.jl
```diff
@@ -518,20 +527,37 @@ end
 # covm (with provided mean)
 ## Use map(t -> t - xmean, x) instead of x .- xmean to allow for Vector{Vector}
 ## which can't be handled by broadcast
+covm(itr::Any, itrmean; corrected::Bool=true) =
+    covm(map(t -> t - itrmean, x); corrected = corrected)
 covm(x::AbstractVector, xmean; corrected::Bool=true) =
```
nalimilan, line 532: This method is identical to the previous one so it's no longer needed.

src/Statistics.jl
```diff
@@ -518,20 +527,37 @@ end
...
+"""
+    cov(x::Any; corrected::Bool=true)
+
+Compute the variance of the iterator `x`. If `corrected` is `true` (the default) then the sum
+is scaled with `n-1`, whereas the sum is scaled with `n` if `corrected` is `false` where `n`
+is the number of elements in the iterator, which is not necessarily known. 
+"""
+function cov(x::Any, corrected::Bool=true)
+    cx = collect(x)
+    covm(cx, mean(cx); corrected=corrected)
```
nalimilan, line 553: This is going to make another copy of `cx`. Better call `covzm` directly.

src/Statistics.jl
```diff
@@ -479,13 +481,19 @@ end
 _vmean(x::AbstractVector, vardim::Int) = mean(x)
 _vmean(x::AbstractMatrix, vardim::Int) = mean(x, dims=vardim)
 
-# core functions
+_abs2(x::Number) = abs2(x)
+_abs2(x)       = x*x'
+
+_conjmul(x::Number, y::Number) = x * conj(y)
```
nalimilan, line 487: Is this still needed?

PR #31, opened 2020-03-18
WIP Progress on incorporating StatsBase into statistics. 

pdeffebach:


pdeffebach:
@nalimilan this is ready for a review. My impression is that not everything has to be 100% perfect for this PR since it's being merged into your PR, and I would bet rebasing there is a high cost of keeping everything up to date. 

The `quantile` function that returns a `Vector` of the quantiles is moved into Statistics, as is `percentile`. Everything else is the same. 

Thanks!

src/Statistics.jl
```diff
@@ -1,5 +1,4 @@
 # This file is a part of Julia. License is MIT: https://julialang.org/license
-
```
nalimilan: Revert this.

src/scalarstats.jl
```diff
@@ -165,24 +165,6 @@ end
 #
 #############################
 
-"""
```
nalimilan: Maybe keep these here for now. That reduces the diff, which makes things less messy, and we may want to move `quantile` to this file (or a separate file) later.

Project.toml
```diff
@@ -1,8 +1,8 @@
 name = "Statistics"
-uuid = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
```
nalimilan: I think you need to keep this to get CI to pass. We'll remove it later.

src/scalarstats.jl
pdeffebach: I tried to do this but can't since this extends methods in statistics and we `include` `scalarstats.jl` at the top.

Project.toml
pdeffebach: I need to delete it in order for a Revise-based workflow to work, see my discourse question [here](https://discourse.julialang.org/t/fork-and-build-julia-for-a-pr-to-stdlib/35926).

pdeffebach:
Sorry I couldn't give better feedback, this is tough. 

Not all tests pass. The `scalarstats` ones, including the `quantile` and `percentile` tests pass. But none of the tests relating to weighted sum and weighted mean pass. At the top of `wsum.jl` there is the `if` statement

```
# Weighted sum is only supported in Julia 1.4 and above
if VERSION >= v"1.4.0"
    _sum = sum
    _sum! = sum!
else
# all the code
```

So it's no surprise that isn't working. Though I don't understand what's going on. Can you help?

nalimilan:
Sorry I missed your question. The `VERSION` check is incorrect as I wrote it before 1.4 was released, thinking that https://github.com/JuliaLang/julia/pull/33310 would be included in that release. So you can just remove it for now and use the code from the other branch.

src/scalarstats.jl
nalimilan: OK. Then remove the heading above.

musm:
Isn't this PR missing a large portion of `StatsBase` ? If I'm not mistaken wasn't the goal to incorporate `StatsBase` into `Statistics`?

nalimilan:
It's marked as WIP. And see also https://github.com/JuliaLang/Statistics.jl/pull/2.

nalimilan:
I've moved the changes from this PR to https://github.com/JuliaLang/Statistics.jl/pull/2.

PR #32, opened 2020-04-05
Add missing shape checks for the means argument to var[m] and std[m]

nalimilan:
We use `@inbounds`, but the shape of the `means` argument was never checked for the general `AbstractArray` method. With an incorrect shape, invalid results or crashes would happen.
To avoid breaking existing code which was working, allow trailing singleton dimensions. Sync the `SparseMatrixCSV` method, which makes it less strict.

See [this Discourse thread](https://discourse.julialang.org/t/bug-in-var-when-supplying-means-as-vectors/36979).

### DIFF
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -235,6 +235,11 @@ centralize_sumabs2(A::AbstractArray, m, ifirst::Int, ilast::Int) =
 function centralize_sumabs2!(R::AbstractArray{S}, A::AbstractArray, means::AbstractArray) where S
     # following the implementation of _mapreducedim! at base/reducedim.jl
     lsiz = Base.check_reducedims(R,A)
+    for i in 1:ndims(R)
+        if axes(means, i) != axes(R, i)
+            throw(DimensionMismatch("dimension $i of `mean` should have indices $(axes(R, i)), but got $(axes(means, i))"))
+        end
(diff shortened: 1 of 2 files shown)


src/Statistics.jl
```diff
@@ -1053,7 +1062,11 @@ end
 function centralize_sumabs2!(R::AbstractArray{S}, A::SparseMatrixCSC{Tv,Ti}, means::AbstractArray) where {S,Tv,Ti}
     require_one_based_indexing(R, A, means)
     lsiz = Base.check_reducedims(R,A)
-    size(means) == size(R) || error("size of means must match size of R")
+    for i in 1:max(ndims(R), ndims(means))
+        if axes(means, i) != axes(R, i)
+            throw(DimensionMismatch("dimension $i of `mean` should have indices $(axes(R, i)), but got $(axes(means, i))"))
+        end
+    end
```
ararslan, line 1069: This code is a little long-ish and is used in two places, so it would make sense to me to factor it into its own little function. Not necessary though.

nalimilan: Well if you leave me the choice... :-p

ararslan: I decree that you are free to do as you wish :stuck_out_tongue:

PR #36, opened 2020-05-08
remove test for empty tuple for mean

KristofferC:
https://github.com/JuliaLang/julia/pull/35803 broke the tests for Statistics because an ambiguity was fixed:

```
julia> using Statistics

julia> mean(())
ERROR: MethodError: reduce_empty(::typeof(Base.add_sum), ::Type{Union{}}) is ambiguous. Candidates:
  reduce_empty(::typeof(Base.add_sum), ::Type{T}) where T<:Union{Int16, Int32, Int8} in Base at reduce.jl:314
  reduce_empty(::typeof(Base.add_sum), ::Type{T}) where T<:Union{UInt16, UInt32, UInt8} in Base at reduce.jl:315
Possible fix, define
  reduce_empty(::typeof(Base.add_sum), ::Type{Union{}})
```
which after that PR is now (more correctly):

```
julia> mean(())
ERROR: ArgumentError: reducing over an empty collection is not allowed
```

This test doesn't seem to do anything so just delete it.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -67,7 +67,6 @@ end
 end
 
 @testset "mean" begin
-    @test_throws MethodError mean(())
     @test mean((1,2,3)) === 2.
     @test mean([0]) === 0.
     @test mean([1.]) === 1.


timholy:
Without that PR it's
```julia
julia> mean(())
ERROR: MethodError: zero(::Type{Union{}}) is ambiguous. Candidates:
  zero(::Type{var"#s822"} where var"#s822"<:AbstractIrrational) in Base at irrationals.jl:149
  zero(::Union{Type{P}, P}) where P<:Dates.Period in Dates at /home/tim/src/julia-master/usr/share/julia/stdlib/v1.6/Dates/src/periods.jl:53
  zero(::Type{T}) where T<:Number in Base at number.jl:242
Possible fix, define
  zero(::Type{Union{}})
Stacktrace:
 [1] reduce_empty(::typeof(+), ::Type{T} where T) at ./reduce.jl:310
 [2] mapreduce_empty(::typeof(identity), ::Function, ::Type{T} where T) at ./reduce.jl:339
 [3] reduce_empty(::Base.MappingRF{typeof(identity),typeof(+)}, ::Type{T} where T) at ./reduce.jl:325
 [4] reduce_empty_iter at ./reduce.jl:351 [inlined]
 [5] mapreduce_empty_iter(::Function, ::Function, ::Tuple{}, ::Base.HasEltype) at ./reduce.jl:347
 [6] mean(::typeof(identity), ::Tuple{}) at /home/tim/src/julia-master/usr/share/julia/stdlib/v1.6/Statistics/src/Statistics.jl:64
 [7] mean(::Tuple{}) at /home/tim/src/julia-master/usr/share/julia/stdlib/v1.6/Statistics/src/Statistics.jl:44
 [8] top-level scope at REPL[3]:1
```
This might be a consequence of
```julia
julia> Base.IteratorEltype(())
Base.HasEltype()
```
I think for Julia 1.6 we should explore
```diff
diff --git a/base/generator.jl b/base/generator.jl
index 7cd075d255..cfe520070f 100644
--- a/base/generator.jl
+++ b/base/generator.jl
@@ -126,3 +126,6 @@ IteratorEltype(::Type) = HasEltype()  # HasEltype is the default
 IteratorEltype(::Type{Generator{I,T}}) where {I,T} = EltypeUnknown()
 
 IteratorEltype(::Type{Any}) = EltypeUnknown()
+
+IteratorEltype(::Type{<:Tuple}) = EltypeUnknown()
+IteratorEltype(::Type{Tuple{T,Vararg{T}}}) where {T} = HasEltype()
```
Running local tests now to see what breaks.

tkf:
:+1: for this change

I think `reduce_empty` should handle `Union{}` correctly. So, a part of JuliaLang/julia#35803 does make sense even only from the point of view of correctness. For example, `sum(Union{}[])` should not throw `MethodError`.

tkf:
This is required for https://github.com/JuliaLang/julia/pull/35843

nalimilan:
Sorry for not responding before, but can you add a test for the new behavior conditional on the version? That sounds like a legitimate case to test.

tkf:
@nalimilan see https://github.com/JuliaLang/Statistics.jl/pull/40

PR #38, opened 2020-05-08
Force specialization for _mean's dims argument

### LINKED ISSUE #37: mean's dims argument is specializing as `::Function` instead of `::Colon`
We use the `:` throughout as a sentinel for "all dimensions."  We do not typically call it as a function in this use case, even though `Colon <: Function`.

However, I think this means that it's getting the specialization rules for `Function`s and thus no longer specializing as it was previously expected to.  In particular, this is introducing type-instabilities that are slowing down basic functions like `mean`:

```Julia
#### master (b72e191ea9) ####
julia> using Statistics, BenchmarkTools

julia> @benchmark mean($(zeros(10)))
BenchmarkTools.Trial:
  memory estimate:  16 bytes
  allocs estimate:  1
  --------------
  minimum time:     11.117 ns (0.00% GC)
  median time:      11.449 ns (0.00% GC)
  mean time:        13.226 ns (1.25% GC)
  maximum time:     635.868 ns (96.44% GC)
  --------------
  samples:          10000
  evals/sample:     998

#### 1.4.0 ####
julia> @benchmark mean($(zeros(10)))
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     4.453 ns (0.00% GC)
  median time:      5.994 ns (0.00% GC)
  mean time:        5.884 ns (0.00% GC)
  maximum time:     37.309 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000
```

(I originally thought this was a Cthulhu bug, but it does appear that `@descend` is properly showing me how Julia is specializing this: https://github.com/JuliaDebug/Cthulhu.jl/issues/77)

mbauman:
Ensures we get the specialization we want when passing `dims=(:)`.  Fixes #37.

### DIFF
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -165,7 +165,7 @@ mean(A::AbstractArray; dims=:) = _mean(identity, A, dims)
 
 _mean_promote(x::T, y::S) where {T,S} = convert(promote_type(T, S), y)
 
-function _mean(f, A::AbstractArray, dims=:)
+function _mean(f, A::AbstractArray, dims::Dims=:) where Dims
     isempty(A) && return sum(f, A, dims=dims)/0
     if dims === (:)
         n = length(A)
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -165,6 +165,7 @@ mean(A::AbstractArray; dims=:) = _mean(identity, A, dims)
 
 _mean_promote(x::T, y::S) where {T,S} = convert(promote_type(T, S), y)
 
+# ::Dims is there to force specializing on Colon (as it is a Function)
 function _mean(f, A::AbstractArray, dims::Dims=:) where Dims
     isempty(A) && return sum(f, A, dims=dims)/0
     if dims === (:)


src/Statistics.jl
```diff
@@ -165,7 +165,7 @@ mean(A::AbstractArray; dims=:) = _mean(identity, A, dims)
 
 _mean_promote(x::T, y::S) where {T,S} = convert(promote_type(T, S), y)
 
-function _mean(f, A::AbstractArray, dims=:)
+function _mean(f, A::AbstractArray, dims::Dims=:) where Dims
```
nalimilan, line 168:
```suggestion
# ::Dims is there to force specializing on Colon (as it is a Function)
function _mean(f, A::AbstractArray, dims::Dims=:) where Dims
```

nalimilan:
Thanks! That's really tricky, it didn't occur to me that `:` was a function.

mbauman:
Yeah, it clearly caught me off-guard, too!  Just look at all the breadcrumbs it took me to get to the correct issue.  Hah!

PR #42, opened 2020-06-01
Fix bug in quantile for case of 1 element vector

bkamins:


bkamins:
The problem is that `clamp` first applies the upper bound condition.

test/runtests.jl
```diff
@@ -557,6 +557,7 @@ end
     @test_throws ArgumentError quantile([1, missing], 0.5)
     @test_throws ArgumentError quantile([1, NaN], 0.5)
     @test quantile(skipmissing([1, missing, 2]), 0.5) === 1.5
+    @test quantile([1]), 0.5) === 1.0
```
bkamins:
```suggestion
    @test quantile([1], 0.5) === 1.0
```

PR #48, opened 2020-07-31
Remove `using Statistics` from jldoctest

zdroid:
Other modules seem not to use this syntax, It was added in #21, but `jldoctest`s should still work without it.

I tried to build the docs locally too (to check if `jldoctest`s work), but `IOError` kept me from finishing that. :/


zdroid:
Running this locally again, it seems to produce errors, so `using ...` is required.

PR #52, opened 2020-09-24
add sample to Statistics

bkamins:
We can have an implementation in StatsBase.jl, but it would be nice that this name is defined in `Statistics` module.

### DIFF
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -1152,7 +1152,7 @@ Uniformly select a single random element of `a`.
 Optionally specify a random number generator `rng` as the first argument
 (defaults to `Random.GLOBAL_RNG`).
 """
-sample(rng::AbstractRNG, a::AbstractArray) = a[rand(rng, eachindex(a))]
+sample(rng::AbstractRNG, a::AbstractArray) = rand(rng, a)
 sample(a::AbstractArray) = sample(Random.GLOBAL_RNG, a)
 
 end # module
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -12,7 +12,7 @@ using LinearAlgebra, SparseArrays
 using Base: has_offset_axes, require_one_based_indexing
 
 export cor, cov, std, stdm, var, varm, mean!, mean,
-    median!, median, middle, quantile!, quantile
+    median!, median, middle, quantile!, quantile, sample
 
 ##### mean #####
 
@@ -1141,5 +1141,9 @@ function centralize_sumabs2!(R::AbstractArray{S}, A::SparseMatrixCSC{Tv,Ti}, mea
     end
     return R
 end
+
+##### Sampling #####
+
+function sample end
 
 end # module
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -7,7 +7,7 @@ Standard library module for basic statistics functionality.
 """
 module Statistics
 
-using LinearAlgebra, SparseArrays
+using LinearAlgebra, Random, SparseArrays
 
 using Base: has_offset_axes, require_one_based_indexing
 
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -1152,7 +1152,7 @@ Uniformly select a single random element of `a`.
 Optionally specify a random number generator `rng` as the first argument
 (defaults to `Random.GLOBAL_RNG`).
 """
-sample(rng::AbstractRNG, a::AbstractArray) = a[rand(rng, 1:length(a))]
+sample(rng::AbstractRNG, a::AbstractArray) = a[rand(rng, eachindex(a))]
 sample(a::AbstractArray) = sample(Random.GLOBAL_RNG, a)
 
 end # module
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -1144,6 +1144,15 @@ end
 
 ##### Sampling #####
 
-function sample end
+"""
+    sample([rng], a)
+
+Uniformly select a single random element of `a`.
+
+Optionally specify a random number generator `rng` as the first argument
+(defaults to `Random.GLOBAL_RNG`).
+"""
+sample(rng::AbstractRNG, a::AbstractArray) = a[rand(rng, 1:length(a))]
+sample(a::AbstractArray) = sample(Random.GLOBAL_RNG, a)
 
 end # module
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,6 +1,6 @@
 # This file is a part of Julia. License is MIT: https://julialang.org/license
 
-using Statistics, Test, Random, LinearAlgebra, SparseArrays, StableRNGs
+using Statistics, Test, Random, LinearAlgebra, SparseArrays
 using Test: guardseed
 
 Random.seed!(123)
@@ -145,7 +145,7 @@ end
     @test mean(Number[1, 1.5, 2+3im]) === 1.5+1im # mixed-type array
     @test mean(v for v in Number[1, 1.5, 2+3im]) === 1.5+1im
     @test (@inferred mean(Int[])) === 0/0
-    @test (@inferred mean(Float32[])) === 0.f0/0    
+    @test (@inferred mean(Float32[])) === 0.f0/0
     @test (@inferred mean(Float64[])) === 0/0
     @test (@inferred mean(Iterators.filter(x -> true, Int[]))) === 0/0
     @test (@inferred mean(Iterators.filter(x -> true, Float32[]))) === 0.f0/0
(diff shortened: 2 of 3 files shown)


mschauer:
But then StatsBase.jl has to be changed in unison - is there already a corresponding PR together with this?

bkamins:
I will open it there when we have a consensus to go this way.

nalimilan:
I don't think we have other cases of empty definitions like this in the stldib. I find it a bit weird, and if we defined a method e.g. in DataFrames, people would get a `MethodError` if they try `sample(1:10)` without `using StatsBase`. I'd rather move the whole function to Statistics.

bkamins:
Then I guess you want to move `sample!` and `wsample` also and all the weighting related types?

nalimilan:
`wsample` is redundant with passing a weights vector as an argument, right? It should probably go away. And the various sampling methods could become types on which to dispatch, and potentially live in StatsBase or in a dedicated package (unless we agree to put them in Statistics).

bkamins:
Well - so it means it probably will not land soon, as what you have described is a great plan (I really love it) but it is quite a big redesign.

nalimilan:
We could at least move `sample` and `sample!` with basic features and keep the rest in StatsBase for now.

mschauer:
It would be ok make Statistics the "product owner" of `sample!` and `sample` and perhaps some other functions through empty definitions before transferring them, of course if we only do it for `sample` and `sample!`, we might transfer the methods right away.

bkamins:
We can easily move this:
```
"""
    sample([rng], a)

Uniformly select a single random element of `a`.

Optionally specify a random number generator `rng` as the first argument
(defaults to `Random.GLOBAL_RNG`).
"""
sample(rng::AbstractRNG, a::AbstractArray) = a[rand(rng, 1:length(a))]
sample(a::AbstractArray) = sample(Random.GLOBAL_RNG, a)
```

For `sample!` there is no easy definition (in general other definitions of `sample` and `sample!` are more advanced - either rely on weights or require more complex algorithms that are defined in StatsBase.jl).

Given this we have two options in my opinion (but please comment if you disagree):
1. Move a larger portion of StatsBase.jl to Statistics (including `AbstractWeights` I think)
2. Leave only "bare" function definitions in Statistics and require loading StatsBase.jl if the user wants more functionality (this is proposed in this PR)

Thinking of the original reason of opening this PR (DataFrames.jl) I think we should pick option 1., but please confirm we go this way before someone starts working on it as it is a major refactoring effort (and requires synchronization between StatsBase.jl and Statistics).

Also I would recommend to make this decision fast before new Julia LTS is shipped (otherwise we will have a nightmare with maintaining StatsBase.jl as it will have to conditionally define the codes depending on the Julia version and be constantly kept in sync with Statistics).

nalimilan:
Yes I think moving the two simple methods makes sense. If you update the PR we can merge it soon, but I'm not sure whether it's still time to add this kind of thing to Julia 1.6 (i.e. update the version of Statistics it uses).

src/Statistics.jl
```diff
@@ -1142,4 +1142,8 @@ function centralize_sumabs2!(R::AbstractArray{S}, A::SparseMatrixCSC{Tv,Ti}, mea
     return R
 end
 
+##### Sampling #####
+
+function sample end
```
bkamins, line 1147:
```suggestion
"""
    sample([rng], a)

Uniformly select a single random element of `a`.

Optionally specify a random number generator `rng` as the first argument
(defaults to `Random.GLOBAL_RNG`).
"""
sample(rng::AbstractRNG, a::AbstractArray) = a[rand(rng, 1:length(a))]
sample(a::AbstractArray) = sample(Random.GLOBAL_RNG, a)
```

src/Statistics.jl
```diff
@@ -1144,6 +1144,15 @@ end
...
-function sample end
+"""
+    sample([rng], a)
+
+Uniformly select a single random element of `a`.
+
+Optionally specify a random number generator `rng` as the first argument
+(defaults to `Random.GLOBAL_RNG`).
+"""
+sample(rng::AbstractRNG, a::AbstractArray) = a[rand(rng, 1:length(a))]
```
mschauer, line 1155:
```suggestion
sample(rng::AbstractRNG, a::AbstractArray) = a[rand(rng, eachindex(a))]
```
Shouldn't this be `eachindex`?

test/runtests.jl
```diff
@@ -857,3 +857,9 @@ end
         @test isfinite.(cov_sparse) == isfinite.(cov_dense)
     end
 end
+
+@testset "sample" begin
+    Random.seed!(1234)
```
mschauer, line 862: This seems brittle as we do not guarantee same samples across architecture.

src/Statistics.jl
bkamins: it should - exactly the same now occurred to me that we have a bug in StatsBase.jl.

test/runtests.jl
mschauer: Reference: https://github.com/JuliaRandom/StableRNGs.jl

bkamins: right - fixed

nalimilan: Are we really allowed to use StableRNGs in the stdlib? I'm afraid not. That not a big deal though since the stdlib follows Julia versions so there's no risk of getting out of sync.

bkamins: We use it only in extras. Why do you think it should not be allowed in tests?

mschauer: Who could we ask?

nalimilan: Maybe first change the Statistics code in your local Julia repo and see whether it builds and passes tests? I don't see how it could work since StableRNGs will be missing.

mschauer: I have a stupid solution:
```
using Random
mutable struct StupidRNG{T} <: Random.AbstractRNG
    i::T
end
function Base.rand(rng::StupidRNG, ::Random.SamplerType{UInt64})   
   rng.i
end
struct StupidRange{T} <: Random.Sampler{T}
   a::T
end
Random.Sampler(::Type{StupidRNG{Int}}, r::AbstractUnitRange{Int}, ::Random.Repetition) = StupidRange(r)
Base.rand(rng::StupidRNG, sp::StupidRange) = mod1(rng.i, length(sp.a))
rand(StupidRNG(3), eachindex([1,2,3,4]))
```

nalimilan: Yeah. But it's weird to have to do this here, I wonder how it's done in Base in general.

Anyway, the tests you added are quite weak, we just assume the hardcoded result is correct but without any way to check it's the case. Maybe better copy tests from StatsBase which check that probabilities are correct, and which should pass even if the RNG changes:
https://github.com/JuliaStats/StatsBase.jl/blob/1c3f3d7ae4fd8afae126443d15ddaf8029316618/test/sampling.jl#L94-L122

nalimilan: Or simpler: just check that the output is equal to its definition (in terms of `rand`). Anyway it's trivial, we mostly rely on `rand`, which is test elsewhere.

bkamins: I have added a test for equality with `rand` and for uniformity. The test for uniformity is only for the default RNG, as in general RNG could be bad and fail the test.

test/runtests.jl
```diff
@@ -859,11 +859,31 @@ end
...
+        mt1 = Random.MersenneTwister(1234)
+        mt2 = Random.MersenneTwister(1234)
+
+        @test [sample(mt1, a) for _ in 1:100] == [rand(mt2, a) for _ in 1:100]
+    end
+
+    @testset "uniformity" begin
+        Random.seed!(1234)
+        counts = zeros(Int, 10)
+        for _ in 1:1_000_000
```
mschauer, line 880: Let's save a bit electricity, also the test will fail with probability > 1%, which becomes prohibitive with more tests of this form. How about
```
     counts = zeros(Int, 5)
      for _ in 1:100_000
          counts[sample(1:5)] += 1
      end
      minimum(counts) > 19_550 # Failure probability ≈ 0.15%
```

src/Statistics.jl
```diff
@@ -1142,4 +1142,17 @@ function centralize_sumabs2!(R::AbstractArray{S}, A::SparseMatrixCSC{Tv,Ti}, mea
...
+
+"""
+    sample([rng], a)
+
+Uniformly select a single random element of `a`.
+
+Optionally specify a random number generator `rng` as the first argument
+(defaults to `Random.GLOBAL_RNG`).
+"""
+sample(rng::AbstractRNG, a::AbstractArray) = a[rand(rng, eachindex(a))]
```
nalimilan, line 1155: Actually why not call `rand` directly on `a`?
```suggestion
sample(rng::AbstractRNG, a::AbstractArray) = rand(rng, a)
```

This stresses even more that without more methods `sample` isn't very useful (apart from defining the function in the stdlib). We should probably move `sample([rng], a, n::Integer; replace=true, ordered=false)` to Statistics soon after this PR.

test/runtests.jl
bkamins: OK

test/runtests.jl
```diff
@@ -857,3 +857,33 @@ end
...
+            counts[sample(1:10)] += 1
+        end
+
+        # 3 standard deviations of the fraction estimator distribution is ~900
+        @test minimum(counts) > 99_000
+        @test maximum(counts) < 101_000
```
bkamins, line 886:
```suggestion
        counts = zeros(Int, 5)
        for _ in 1:100_000
            counts[sample(1:5)] += 1
        end

        # violation probability is <0.001
        @test minimum(counts) > 19_500
        @test maximum(counts) < 20_500
```

bkamins:
@ararslan - after we merge this what is the recommended path we should make to sync it with:
1. Compat.jl
2. StatsBase.jl

so that e.g. DataFrames.jl can cleanly add a method for `sample` later?

nalimilan:
If we want to overload this in DataFrames, better put it in Compat.jl.

ararslan:
> If we want to overload this in DataFrames, better put it in Compat.jl.

Does DataFrames depend on Statistics? Putting it in Compat seems right as a first step, and it can be extended from Compat in StatsBase as well.

Aside from that though, I'm wondering if Random doesn't make more sense as a home for `sample`, then extend and re-export from Statistics with methods added for weights, etc.? Currently it seems `sample` is effectively an alias for `rand`.

nalimilan:
Yeah `sample` could also live in Random. Though it would be simpler for developers and users (notably for documentation) to have all methods in the same module. So if we don't want to put weighted methods in Random I'd rather put everything in Statistics. Note that weighted methods don't have to depend on `AbstractWeights` I think, since all weights have the same effect (contrary to e.g. the correction for `var`). So they could live in Random, taking any `AbstractVector`.

nalimilan:
The definition should probably live in StatsAPI now that it exists.

bkamins:
Closing it as after thinking if someone defines `sample` then StatsBase.jl is needed anyway.

Issue #53, opened 2020-09-25
Cannot compute quantile(x::Array{Millisecond,1})


ianfiske:
```
ms = Millisecond.(1:10)
quantile(ms, 0.5)
```
gives

```
ERROR: MethodError: no method matching isnan(::Millisecond)
Closest candidates are:
  isnan(::BigFloat) at mpfr.jl:881
  isnan(::Missing) at missing.jl:100
  isnan(::Float16) at float.jl:537
  ...
Stacktrace:
 [1] _quantilesort!(::Array{Millisecond,1}, ::Bool, ::Float64, ::Float64) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/Statistics/src/Statistics.jl:960
 [2] quantile!(::Array{Millisecond,1}, ::Float64; sorted::Bool, alpha::Float64, beta::Float64) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/Statistics/src/Statistics.jl:943
 [3] quantile(::Array{Millisecond,1}, ::Float64; sorted::Bool, alpha::Float64, beta::Float64) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/Statistics/src/Statistics.jl:1052
 [4] quantile(::Array{Millisecond,1}, ::Float64) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/Statistics/src/Statistics.jl:1052
 [5] top-level scope at REPL[14]:1
 [6] include_string(::Function, ::Module, ::String, ::String) at ./loading.jl:1088
```

on Julia 1.5.2.

The root issue is of course 
```
julia> isnan(Millisecond(1))
ERROR: MethodError: no method matching isnan(::Millisecond)
Closest candidates are:
  isnan(::BigFloat) at mpfr.jl:881
  isnan(::Missing) at missing.jl:100
  isnan(::Float16) at float.jl:537
```

since `_quantilesort!` calls `isnan(x[end])`

ianfiske:
Would

```
isnan(::TimePeriod) = false
```

be an acceptable fix?

mschauer:
Unitful defines `isnan` https://github.com/PainterQubits/Unitful.jl/blob/2be7e0692cafd948b66132920ee407b73220717f/src/quantities.jl
and base should perhaps define it as well.

[...]

PR #54, opened 2020-09-29
Only check isnan if applicable

ianfiske:
Fix https://github.com/JuliaLang/Statistics.jl/issues/53

### DIFF
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -957,7 +957,7 @@ function _quantilesort!(v::AbstractArray, sorted::Bool, minp::Real, maxp::Real)
         sort!(v, 1, lv, Base.Sort.PartialQuickSort(lo:hi), Base.Sort.Forward)
     end
     ismissing(v[end]) && throw(ArgumentError("quantiles are undefined in presence of missing values"))
-    isnan(v[end]) && throw(ArgumentError("quantiles are undefined in presence of NaNs"))
+    applicable(isnan, v[end]) && isnan(v[end]) && throw(ArgumentError("quantiles are undefined in presence of NaNs"))
     return v
 end
 
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -2,6 +2,7 @@
 
 using Statistics, Test, Random, LinearAlgebra, SparseArrays
 using Test: guardseed
+using Dates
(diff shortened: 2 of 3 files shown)


ianfiske:
As per the discussion in #53, I added a check if `isnan` is applicable before calling it.  Note that now, the tests fail because there is also no `isfinite` method for `Millisecond`.  I can fix this with the same strategy of checking if the method is `applicable`, but this just feels clunky to me.  I'm used to the "call the method" and it does the right thing for each type.  Checking before calling so many methods just starts to look clunky, but if you think that's the right approach, I can easily add it.

nalimilan:
I agree it's not ideal, but the only other solution I can see is to define `isnan(::Any)` and `isfinite(::Any)` to be `false`. Do you feel like proposing this in Base? That would probably allow making the code cleaner.

CI errors, but AFAICT it's no related to `isfinite`?

nalimilan:
Bump.

ianfiske:
@nalimilan Thanks for the bump, I've been busy with other things and I don't think I have time to make the PR to Base right now.  Feel free to run with it if you have time before I do.

test/runtests.jl
```diff
@@ -695,6 +696,8 @@ end
     @test quantile(skipmissing([1, missing, 2]), 0.5) === 1.5
     @test quantile([1], 0.5) === 1.0
 
+    @test quantile(Millisecond.(1:10), 0.5) == Millisecond(5)
```
nalimilan, line 699:
```suggestion
    @test quantile(Millisecond.(0:10), 0.5) == Millisecond(5)
```

src/Statistics.jl
```diff
@@ -1049,7 +1049,7 @@ end
 
     # When a ≉ b, b-a may overflow
     # When a ≈ b, (1-γ)*a + γ*b may not be increasing with γ due to rounding
-    if isfinite(a) && isfinite(b) &&
+    if applicable(isfinite, a) && isfinite(a) && isfinite(b) &&
```
nalimilan, line 1052:
```suggestion
    if applicable(isfinite, a) && isfinite(a) &&
        applicable(isfinite, b) && isfinite(b)
```

PR #55, opened 2020-10-07
Add a README with instructions for developing the package locally

DilumAluthge:
The Pkg.jl repository has a [README](https://github.com/JuliaLang/Pkg.jl/blob/master/README.md) that includes instructions for developing Pkg.jl locally.

This pull request adds a README with similar instructions for developing Statistics.jl locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,14 @@
+# Statistics.jl
+
+Development repository for the Statistics standard library (stdlib) that ships with Julia. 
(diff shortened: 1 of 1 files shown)


DilumAluthge:
@KristofferC Can you take a look at this and make sure that it looks right (since you wrote the corresponding content in the Pkg.jl README)?

PR #58, opened 2020-12-10
Clarify "IID"

BioTurboNick:
The description of the functions might be overly technical for most users, but still within reach. "IID", however, is opaque if you aren't familiar with the acronym. Spelling it out provides a stronger sense of what is meant to laypeople, IMO.

### DIFF
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -306,8 +306,9 @@ end
 Compute the sample variance of collection `itr`, with known mean(s) `mean`.
 
 The algorithm returns an estimator of the generative distribution's variance
-under the assumption that each entry of `itr` is an IID drawn from that generative
-distribution. For arrays, this computation is equivalent to calculating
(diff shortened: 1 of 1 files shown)


src/Statistics.jl
```diff
@@ -343,8 +344,9 @@ end
 Compute the sample variance of collection `itr`.
 
 The algorithm returns an estimator of the generative distribution's variance
-under the assumption that each entry of `itr` is an IID drawn from that generative
-distribution. For arrays, this computation is equivalent to calculating
+under the assumption that each entry of `itr` drawn from that generative
+distribution is independent and identically distributed (IID).
```
mschauer, line 348: How about
"each entry of `its` is a sample drawn from the same unknown distribution with the samples being uncorrelated."

(also because uncorrelated is enough for this to hold, independence is not needed in the first place)

mschauer (approved):

nalimilan:
Thanks!

PR #61, opened 2021-01-03
change real to float in cor of a single collection

bkamins:
As in `cor` we get square root I think it is safe to assume that the result should be floating point.

An example of current surprising behavior:
```
julia> cor([im])
true
```

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -516,6 +516,20 @@ end
         @test cor(tmp, tmp) <= 1.0
         @test cor(tmp, tmp2) <= 1.0
     end
+    
+    @test cor(Int[]) === 1.0
+    @test cor([im]) ===1.0 + 0.0im
+    @test_throws MethodError cor([])
+    @test_throws MethodError cor(Any[1.0])
+    
+    @test cor([1, missing]) === 1.0
+    @test ismissing(cor([missing]))
+    @test_throws MethodError cor(Any[1.0, missing])
+    
+    @test Statistics.corm([true], 1.0) === 1.0
+    @test_throws MethodError Statistics.corm(Any[0.0, 1.0], 0.5)
+    @test Statistics.corzm([true]) === 1.0
+    @test_throws MethodError Statistics.corzm(Any[0.0, 1.0])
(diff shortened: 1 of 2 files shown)


nalimilan (approved):
Thanks!

nalimilan:
Though could you add a test to check the new behavior?

bkamins:
A good point - I have added the tests. They highlight the cases that I think are critical. A check that in all cases we produce what we think should be produced is welcome (I think they are not always intuitive, but I made the PR to be non-breaking in these corner cases except changing the type of `1` produced).

test/runtests.jl
```diff
@@ -516,6 +516,20 @@ end
         @test cor(tmp, tmp) <= 1.0
         @test cor(tmp, tmp2) <= 1.0
     end
+    
+    @test cor(Int[]) === 1.0
+    @test cor([im]) ===1.0 + 0.0im
```
nalimilan, line 521:
```suggestion
    @test cor([im]) === 1.0 + 0.0im
```

nalimilan:
I just noticed `corm` and `corzm` have similar definitions. I guess these should be changed too, right?

nalimilan:
@stevengj You introduced this definition at https://github.com/JuliaLang/julia/pull/13348, replacing `float(one(T) * one(T))` with `real(one(T))`. Yet `cor([1im, 2im], [1im, 2im])` returns `1.0 + 0.0im`, not `1.0`. I guess nobody really tested nor care about this use case. What do you think we should do? We could also use `float(real(one(T))` to minimize breakage; what matters is that `cor([1, 2])` returns `1.0`, not `1`.

bkamins:
> I guess these should be changed too, right?

What do you mean exactly? I have already changed `corm` and `corzm` in this PR (and added tests for this).

nalimilan:
> What do you mean exactly? I have already changed `corm` and `corzm` in this PR (and added tests for this).

Woops, forget it, I was just confused.

nalimilan:
@stevengj Any chance you could find the time to look at this?

PR #70, opened 2021-01-23
Allow passing 0-length vectors to cor

bkamins:
I have not added the tests yet. Let us first decide if we like the design of what I propose and then I will finalize the PR.

The proposed change is non-breaking.

### DIFF
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -685,7 +685,14 @@ function corm(x::AbstractVector, mx, y::AbstractVector, my)
     require_one_based_indexing(x, y)
     n = length(x)
     length(y) == n || throw(DimensionMismatch("inconsistent lengths"))
-    n > 0 || throw(ArgumentError("correlation only defined for non-empty vectors"))
+    if n == 0
+        T = promote_type(typeof(mx), typeof(my))
+        if T <: AbstractFloat
+            return T(NaN)
+        else
+            return NaN
+        end
+    end
 
     @inbounds begin
         # Initialize the accumulators


src/Statistics.jl
```diff
@@ -685,7 +685,14 @@ function corm(x::AbstractVector, mx, y::AbstractVector, my)
     require_one_based_indexing(x, y)
     n = length(x)
     length(y) == n || throw(DimensionMismatch("inconsistent lengths"))
-    n > 0 || throw(ArgumentError("correlation only defined for non-empty vectors"))
+    if n == 0
+        T = promote_type(typeof(mx), typeof(my))
```
nalimilan, line 689: How about using an equation closer to the actual computation of the correlation, like the one used below with `zero`?

bkamins: I did this this way as `x` and `y` can have abstract `eltype` (e.g. `Any`), in which case you cannot use `zero` on it.

This code assumes that type of `mx` and `my` is somehow "derived" from `x` and `y`. As `corm` is unexported I thought it should be safe.

But we might use something like: `sqrt(zero(mx)*zero(my)/(abs2(one(mx)*abs2(one(my)))`

bkamins:
Also probably `cov` should be fixed (I will try to look into it later, after I am done with DataFrames.jl reviews).

nalimilan:
+1

Relevant previous discussions:
- https://github.com/JuliaLang/julia/issues/28777 about `mean` for iterators with known eltype (can this be closed BTW?)
- https://github.com/JuliaLang/julia/issues/19444 about `median`
- https://github.com/JuliaLang/julia/issues/5234 is more general but seems to have focused mostly on fast scalar functions

src/Statistics.jl
nalimilan: Yeah, that formula looks good (with `zero` at the denominator)!

PR #72, opened 2021-02-13
Fix NaN and missing detection in quantile()

nalimilan:
When `sort=false`, we only partially sort the input, so `NaN`/`missing` is not guaranteed to be in the last position. Also avoid throwing errors for non-`Number` types, for which `isnan` may not be defined.

Partially supersedes https://github.com/JuliaLang/Statistics.jl/pull/54/. Relies on https://github.com/JuliaLang/julia/pull/38232 to ensure that `isnan` works for any type (but before things were even worse).


mkborregaard (approved):
great

PR #74, opened 2021-02-13
fix handling of Missing in cor

bkamins:


src/Statistics.jl
```diff
@@ -713,7 +715,8 @@ corm(x::AbstractVecOrMat, xmean, y::AbstractVecOrMat, ymean, vardim::Int=1) =
 
 Return the number one.
 """
-cor(x::AbstractVector) = one(float(eltype(x)))
+cor(x::AbstractVector) =
```
nalimilan, line 718:
```suggestion
cor(x::AbstractVector) where {T} =
```

src/Statistics.jl
```diff
@@ -713,7 +715,8 @@ corm(x::AbstractVecOrMat, xmean, y::AbstractVecOrMat, ymean, vardim::Int=1) =
 
 Return the number one.
 """
-cor(x::AbstractVector) = one(float(eltype(x)))
+cor(x::AbstractVector) where {T} =
```
nalimilan, line 718:
```suggestion
cor(x::AbstractVector{T}) where {T} =
```

src/Statistics.jl
```diff
@@ -713,7 +715,8 @@ corm(x::AbstractVecOrMat, xmean, y::AbstractVecOrMat, ymean, vardim::Int=1) =
...
+    one(float(eltype(T === Missing ? Missing : nonmissingtype(T))))
```
nalimilan, line 719:
```suggestion
    one(float(T === Missing ? Missing : nonmissingtype(T)))
```

src/Statistics.jl
```diff
@@ -665,7 +665,8 @@ end
 
 # corzm (non-exported, with centered data)
 
-corzm(x::AbstractVector{T}) where {T} = one(float(T))
+corzm(x::AbstractVector{T}) where {T} =
+    one(float(T === Missing ? Missing : nonmissingtype(T)))
```
nalimilan, line 669: Actually shouldn't this be `one(T === Missing ? Missing : float(nonmissingtype(T)))`? `float(Missing)` doesn't work on Julia 1.6 and below.

bkamins, line 669:
```suggestion
    T === Missing ? missingissing : one(float(nonmissingtype(T)))
```

src/Statistics.jl
```diff
@@ -679,7 +680,8 @@ corzm(x::AbstractMatrix, y::AbstractMatrix, vardim::Int=1) =
 
 # corm
 
-corm(x::AbstractVector{T}, xmean) where {T} = one(float(T))
+corm(x::AbstractVector{T}, xmean) where {T} =
+    one(float(T === Missing ? Missing : nonmissingtype(T)))
```
bkamins, line 684:
```suggestion
    T === Missing ? missing : one(float(nonmissingtype(T)))
```

src/Statistics.jl
```diff
@@ -713,7 +715,8 @@ corm(x::AbstractVecOrMat, xmean, y::AbstractVecOrMat, ymean, vardim::Int=1) =
 
 Return the number one.
 """
-cor(x::AbstractVector) = one(float(eltype(x)))
+cor(x::AbstractVector{T}) where {T} =
+    one(float(T === Missing ? Missing : nonmissingtype(T)))
```
bkamins, line 719:
```suggestion
    T === Missing ? missing : one(float(nonmissingtype(T)))
```

src/Statistics.jl
```diff
@@ -665,7 +665,8 @@ end
 
 # corzm (non-exported, with centered data)
 
-corzm(x::AbstractVector{T}) where {T} = one(float(T))
+corzm(x::AbstractVector{T}) where {T} =
```
bkamins, line 668:
```suggestion
if !isdefined(Base, :nonmissingtype)
    nonmissingtype(::Type{T}) where {T} = Core.Compiler.typesubtract(T, Missing)
end

corzm(x::AbstractVector{T}) where {T} =
```

bkamins:
So a tension in this PR that we have is:
```
julia> using Statistics

julia> z = [1.0, im]
2-element Vector{ComplexF64}:
 1.0 + 0.0im
 0.0 + 1.0im

julia> var(z)
1.0

julia> cov(z, z)
1.0 + 0.0im
```
and we have a tension that the identity `var(z) === cov(z, z)` does not hold. But probably we should accept it.

nalimilan:
Yeah, but AFAICT that's an expected result since the variance is guaranteed to be a real number, contrary to the covariance:
https://en.wikipedia.org/wiki/Complex_random_variable#Variance_and_pseudo-variance
https://selipot.github.io/talks/lecture2.pdf

Anyway `var(z) == cov(z, z)` holds so that's already quite consistent.

src/Statistics.jl
```diff
@@ -665,7 +665,12 @@ end
 
 # corzm (non-exported, with centered data)
 
-corzm(x::AbstractVector{T}) where {T} = one(float(T))
+if !isdefined(Base, :nonmissingtype)
+    nonmissingtype(::Type{T}) where {T} = Core.Compiler.typesubtract(T, Missing)
+end
```
nalimilan, line 670: Was this actually needed for CI to pass? We don't support older Julia versions as Statistics cannot be installed there anyway. Actually we could almost drop support for Julia 1.6 if we needed since this code will only be used in Julia 1.7. :-p

bkamins:
> an expected result since the variance is guaranteed to be a real number

Yes - that is why I said we should accept it.

> Was this actually needed for CI to pass?

Ah - I added it to make the tests pass on 1.0. I will remove it

src/Statistics.jl
```diff
@@ -665,7 +665,12 @@ end
...
+
```
bkamins, line 671:
```suggestion
```

PR #81, opened 2021-07-19
Fix missing `stdm` method with dims argument

ParadaCarleton:
`stdm` is missing a "dims" argument which is documented but has not actually been added.

### DIFF
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -419,8 +419,8 @@ function sqrt!(A::AbstractArray)
     A
 end
 
-stdm(A::AbstractArray, m; corrected::Bool=true) =
-    sqrt.(varm(A, m; corrected=corrected))
+stdm(A::AbstractArray, m; corrected::Bool=true, dims=:) =
+    sqrt.(varm(A, m; corrected=corrected, dims=dims))
 
 """
     std(itr; corrected::Bool=true, mean=nothing[, dims])
@@ -467,7 +467,7 @@ std(iterable; corrected::Bool=true, mean=nothing) =
     sqrt(var(iterable, corrected=corrected, mean=mean))
 
 """
-    stdm(itr, mean; corrected::Bool=true)
+    stdm(itr, mean; corrected::Bool=true, dims)
 
 Compute the sample standard deviation of collection `itr`, with known mean(s) `mean`.
 
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -487,14 +487,11 @@ over dimensions. In that case, `mean` must be an array with the same shape as
     Use the [`skipmissing`](@ref) function to omit `missing` entries and compute the
     standard deviation of non-missing values.
 """
-stdm(A::AbstractArray, m; corrected::Bool=true) =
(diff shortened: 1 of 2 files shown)


nalimilan:
Thanks and sorry for the delay. Could you add tests for the new argument? You can probably copy those used for `stdm`.

ParadaCarleton:
> Thanks and sorry for the delay. Could you add tests for the new argument? You can probably copy those used for `stdm`.

No worries, I also forgot about this :sweat_smile:

ParadaCarleton:
Thanks a bunch to @burmecia for adding tests!

src/Statistics.jl
```diff
@@ -467,7 +464,7 @@ std(iterable; corrected::Bool=true, mean=nothing) =
     sqrt(var(iterable, corrected=corrected, mean=mean))
 
 """
-    stdm(itr, mean; corrected::Bool=true)
+    stdm(itr, mean; corrected::Bool=true, dims)
```
nalimilan, line 467:
```suggestion
    stdm(itr, mean; corrected::Bool=true[, dims])
```

src/Statistics.jl
```diff
@@ -490,6 +487,12 @@ over dimensions. In that case, `mean` must be an array with the same shape as
...
     standard deviation of non-missing values.
 """
+stdm(A::AbstractArray, m; corrected::Bool=true) =
+    sqrt.(varm(A, m; corrected=corrected))
+
+stdm(A::AbstractArray, m::AbstractArray; corrected::Bool=true, dims=:) =
+    sqrt.(varm(A, m; corrected=corrected, dims=dims))
+
 stdm(iterable, m; corrected::Bool=true) =
     std(iterable, corrected=corrected, mean=m)
```
nalimilan, line 497: Let's use the same approach as above for `std`?
```suggestion
stdm(A::AbstractArray, m::AbstractArray; corrected::Bool=true, dims=:) =
    _std(A, corrected, m, dims)

stdm(iterable, m; corrected::Bool=true) =
    sqrt(var(iterable, corrected=corrected, mean=mean))
```

ParadaCarleton: Welp, looks like that broke something. @nalimilan

src/Statistics.jl
```diff
@@ -490,8 +487,11 @@ over dimensions. In that case, `mean` must be an array with the same shape as
     Use the [`skipmissing`](@ref) function to omit `missing` entries and compute the
     standard deviation of non-missing values.
 """
+stdm(A::AbstractArray, m::AbstractArray; corrected::Bool=true, dims=:) =
+    _std(A, corrected, m, dims)
+
 stdm(iterable, m; corrected::Bool=true) =
-    std(iterable, corrected=corrected, mean=m)
+    sqrt(var(iterable, corrected=corrected, mean=mean))
```
nalimilan, line 494: Woops, should have been:
```suggestion
    sqrt(var(iterable, corrected=corrected, mean=m))
```

nalimilan:
Thanks!

PR #85, opened 2021-09-07
Improve numerical stability of `cov`

nalimilan:
`cov(::AbstractVector, ::AbstractVector)` was less stable than other `cov` methods and than `var` on arrays as it called `sum` on a generator, which does not use pairwise summation. Use a `Broadcasted` object instead to benefit from pairwise summation.

Fixes https://github.com/JuliaLang/Statistics.jl/issues/83.

Performance also seems to increase a bit:
```julia
# Before
julia> A = 20*randn(Float32, 10_000_000) .+ 100;

julia> @btime cov(A, A);
  42.300 ms (5 allocations: 76.29 MiB)

# After
julia> @btime cov(A, A);
  33.832 ms (5 allocations: 76.29 MiB)
```

### DIFF
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -519,7 +519,8 @@ unscaled_covzm(x::AbstractVector{<:Number})    = sum(abs2, x)
 unscaled_covzm(x::AbstractVector)              = sum(t -> t*t', x)
 unscaled_covzm(x::AbstractMatrix, vardim::Int) = (vardim == 1 ? _conj(x'x) : x * x')
 
-unscaled_covzm(x::AbstractVector, y::AbstractVector) = sum(conj(y[i])*x[i] for i in eachindex(y, x))
+unscaled_covzm(x::AbstractVector, y::AbstractVector) =
+    sum(Broadcast.instantiate(Broadcast.broadcasted((xi, yi) -> conj(yi)*xi, x, y)))
 unscaled_covzm(x::AbstractVector, y::AbstractMatrix, vardim::Int) =
     (vardim == 1 ? *(transpose(x), _conj(y)) : *(transpose(x), transpose(_conj(y))))
 unscaled_covzm(x::AbstractMatrix, y::AbstractVector, vardim::Int) =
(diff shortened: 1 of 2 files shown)


src/Statistics.jl
```diff
@@ -519,7 +519,8 @@ unscaled_covzm(x::AbstractVector{<:Number})    = sum(abs2, x)
 unscaled_covzm(x::AbstractVector)              = sum(t -> t*t', x)
 unscaled_covzm(x::AbstractMatrix, vardim::Int) = (vardim == 1 ? _conj(x'x) : x * x')
 
-unscaled_covzm(x::AbstractVector, y::AbstractVector) = sum(conj(y[i])*x[i] for i in eachindex(y, x))
+unscaled_covzm(x::AbstractVector, y::AbstractVector) =
+    sum(Broadcast.instantiate(Broadcast.broadcasted((xi, yi) -> conj(yi)*xi, x, y)))
```
ararslan, line 523: I wonder how this would compare to e.g.
```julia
sum(((xi, yi),) -> conj(yi)*xi, zip(x, y))
```

ararslan (approved):

src/Statistics.jl
nalimilan: That wouldn't fix the bug, as `zip(x, y)` is just an iterator so pairwise summation isn't used. Another possibility would be to use `y'*x`, which appears to be quite accurate too (depending on the cases, more or less than pairwise summation; but that depends on BLAS I guess).
```julia
julia> x = rand(Float32, 10_000);

julia> y = rand(Float32, 10_000);

julia> sum(((xi, yi),) -> conj(yi)*xi, zip(x, y))
2536.1172f0

julia> sum(Broadcast.instantiate(Broadcast.broadcasted((xi, yi) -> conj(yi)*xi, x, y)))
2536.1304f0

julia> y'*x
2536.1306f0

julia> sum(((xi, yi),) -> conj(yi)*xi, zip(big.(x), big.(y)))
2536.130404875756170213207951746881008148193359375
```

ararslan: `y'*x` seems like a good way to do it. Benchmarking suggests it's also fairly significantly faster.

nalimilan: Performance doesn't really matter here as the rest of the code is quite slow (we make two additional passes over the data, including a copy). But yeah it sounds simpler, even though I'm not completely sure what accuracy guarantees `y'*x` provides.

nalimilan: Actually after (painfully) digging in the history of this file, it turns out this was originally what we did, but the call was changed to avoid depending on LinearAlgebra. Know we can go back to it since we're an stdlib!
https://github.com/JuliaLang/julia/pull/25571#discussion_r162214400

PR #86, opened 2021-09-08
Improve quicksort performance

bkamins:
Fixes https://github.com/JuliaLang/Statistics.jl/issues/84

### DIFF
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -938,7 +938,7 @@ function quantile!(q::AbstractArray, v::AbstractVector, p::AbstractArray;
     isempty(q) && return q
 
     minp, maxp = extrema(p)
-    _quantilesort!(v, sorted, minp, maxp)
+    _quantilesort!(v, sorted, minp, maxp, length(p))
 
     for (i, j) in zip(eachindex(p), eachindex(q))
         @inbounds q[j] = _quantile(v,p[i], alpha=alpha, beta=beta)
@@ -950,16 +950,16 @@ function quantile!(v::AbstractVector, p::Union{AbstractArray, Tuple{Vararg{Real}
                    sorted::Bool=false, alpha::Real=1., beta::Real=alpha)
     if !isempty(p)
         minp, maxp = extrema(p)
-        _quantilesort!(v, sorted, minp, maxp)
+        _quantilesort!(v, sorted, minp, maxp, length(p))
     end
     return map(x->_quantile(v, x, alpha=alpha, beta=beta), p)
 end
 
 quantile!(v::AbstractVector, p::Real; sorted::Bool=false, alpha::Real=1., beta::Real=alpha) =
-    _quantile(_quantilesort!(v, sorted, p, p), p, alpha=alpha, beta=beta)
+    _quantile(_quantilesort!(v, sorted, p, p, 1), p, alpha=alpha, beta=beta)
 
 # Function to perform partial sort of v for quantiles in given range
-function _quantilesort!(v::AbstractArray, sorted::Bool, minp::Real, maxp::Real)
+function _quantilesort!(v::AbstractArray, sorted::Bool, minp::Real, maxp::Real, len::Int)
     isempty(v) && throw(ArgumentError("empty data vector"))
     require_one_based_indexing(v)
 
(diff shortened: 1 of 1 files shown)


bkamins:
I still need to add correctness tests.

bkamins:
@nalimilan - doing an "additional sorting" is not correct unfortunately (the second sorting destroys the first), so we need to do this independently unfortunately (still it should not be that bad).

src/Statistics.jl
```diff
@@ -938,7 +938,7 @@ function quantile!(q::AbstractArray, v::AbstractVector, p::AbstractArray;
     isempty(q) && return q
 
     minp, maxp = extrema(p)
-    _quantilesort!(v, sorted, minp, maxp)
+    _quantilesort!(v, sorted, minp, maxp, length(p))
```
bkamins, line 941:
```suggestion
    _quantilesort!(v, sorted, minp, maxp)
```

src/Statistics.jl
```diff
@@ -950,26 +950,34 @@ function quantile!(v::AbstractVector, p::Union{AbstractArray, Tuple{Vararg{Real}
                    sorted::Bool=false, alpha::Real=1., beta::Real=alpha)
     if !isempty(p)
         minp, maxp = extrema(p)
-        _quantilesort!(v, sorted, minp, maxp)
+        _quantilesort!(v, sorted, minp, maxp, length(p))
```
bkamins, line 953:
```suggestion
        _quantilesort!(v, sorted, minp, maxp)
```

src/Statistics.jl
```diff
@@ -950,26 +950,34 @@ function quantile!(v::AbstractVector, p::Union{AbstractArray, Tuple{Vararg{Real}
...
     end
     return map(x->_quantile(v, x, alpha=alpha, beta=beta), p)
 end
 
 quantile!(v::AbstractVector, p::Real; sorted::Bool=false, alpha::Real=1., beta::Real=alpha) =
-    _quantile(_quantilesort!(v, sorted, p, p), p, alpha=alpha, beta=beta)
+    _quantile(_quantilesort!(v, sorted, p, p, 1), p, alpha=alpha, beta=beta)
```
bkamins, line 959:
```suggestion
    _quantile(_quantilesort!(v, sorted, p, p), p, alpha=alpha, beta=beta)
```

src/Statistics.jl
```diff
@@ -950,26 +950,34 @@ function quantile!(v::AbstractVector, p::Union{AbstractArray, Tuple{Vararg{Real}
...
 
 # Function to perform partial sort of v for quantiles in given range
-function _quantilesort!(v::AbstractArray, sorted::Bool, minp::Real, maxp::Real)
+function _quantilesort!(v::AbstractArray, sorted::Bool, minp::Real, maxp::Real, len::Int)
```
bkamins, line 962:
```suggestion
function _quantilesort!(v::AbstractArray, sorted::Bool, minp::Real, maxp::Real)
```

src/Statistics.jl
```diff
@@ -950,26 +950,34 @@ function quantile!(v::AbstractVector, p::Union{AbstractArray, Tuple{Vararg{Real}
...
+            lo2 = ceil(Int,1+minp*(lv))
+            sort!(v, 1, lv, Base.Sort.PartialQuickSort(lo1:lo2), Base.Sort.Forward)
+            hi1 = floor(Int,maxp*(lv))
+            hi2 = ceil(Int,1+maxp*(lv))
+            sort!(v, 1, lv, Base.Sort.PartialQuickSort(hi1:hi2), Base.Sort.Forward)
+        else
+            lo = floor(Int,minp*(lv))
+            hi = ceil(Int,1+maxp*(lv))
+            sort!(v, 1, lv, Base.Sort.PartialQuickSort(lo:hi), Base.Sort.Forward)
+        end
```
bkamins, line 980:
```suggestion
        lo = floor(Int,minp*(lv))
        hi = ceil(Int,1+maxp*(lv))

        # only need to perform partial sort
        sort!(v, 1, lv, Base.Sort.PartialQuickSort(lo:hi), Base.Sort.Forward)        end
```

src/Statistics.jl
```diff
@@ -937,22 +937,33 @@ function quantile!(q::AbstractArray, v::AbstractVector, p::AbstractArray;
     end
     isempty(q) && return q
 
-    minp, maxp = extrema(p)
-    _quantilesort!(v, sorted, minp, maxp)
+    if length(p) == 2
```
nalimilan, line 940: Special-casing 2 is kind of weird. For example, for `[0.25, 0.5, 0.75]` this branch would probably also be faster, right? Actually, isn't this approach faster than the other in most cases?

A possible optimization would be to call `partialsort!` on the full array for the first quantile, then call it on a view from the first quantile to the end of the array for the second, and so on. Not sure that would make a big difference, but at least it shouldn't be really slower than sorting everything between the two extreme quantiles, right?

bkamins: Yes, it is faster:
```
julia> using Statistics

julia> using BenchmarkTools

julia> x = rand(10^4);

julia> @btime quantile($x, [0.25, 0.5, 0.75]);
  410.400 μs (4 allocations: 78.42 KiB)

julia> @btime [quantile($x, 0.25), quantile($x, 0.5), quantile($x, 0.75)];
  314.000 μs (7 allocations: 234.72 KiB)
```
The problem is that quantiles do not need to be sorted, so this complicates the code (but of course is doable as I guess sorting requested quantiles should not be problematic for performance).

We do not need views I think, as `sort!` supports passing start and stop ranges.

However, as it seems to be a bigger rework I will leave it for after DataFrames.jl 1.3 releaes when we work on general Statistics update.

nalimilan: OK. Yes, sorting quantiles should have a negligible cost (and most of the time they will already be sorted).

PR #91, opened 2021-10-17
Improve quantile performance v2

bkamins:
This is an alternative implementation to #86.

Here following https://github.com/JuliaStats/Statistics.jl/pull/86#discussion_r704274221 I perform partial sorting incrementally.

I make this a separate PR to allow an easy comparison of both. Either one or the other should be merged.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -603,7 +603,7 @@ end
 
     # randomized partialsort correctness test
     Random.seed!(1234)
-    for i in 1:100, j in 1:20
+    for i in 1:200, j in 1:20
         x = rand(2000)
         p = rand(j)
         @test quantile(x, p) == [quantile(x, v) for v in p]


nalimilan:
Thanks. Is there any reason to think that a series of partial sorts of nested subsets of the data would be significantly slower than a single full sort? Have you tried benchmarking this?

nalimilan (commented):
I realize I had two comments pending...

nalimilan:
I'm not sure it's worth worrying about performance when the number of quantiles is large compared to the data. Quantiles don't make a lot of sense in that case.

Maybe a simple optimization is to do `sort_p = issorted(p) ? p : sort(p)` to avoid making a copy if possible (as that will be the case almost all the time).

nalimilan:
Actually with the (semi-)recent improvements to sorting performance, it doesn't seem that partial sort is a good idea, as it uses quick sort, while e.g. radix sort is much faster for standard numeric types. Even for `BigInt` sorting everything seems much faster.
 
Maybe we should just switch to a full sort?

I'm a bit surprised that the implementation from this PR is so slow in my benchmark. I would be good to double-check the result. (I think yours had a bug because it reused `x` after sorting it.)

```julia
julia> function f1(n)
           @btime Statistics._quantilesort!($(rand(10^6)), false, extrema($(rand(n)))...)
           @btime new_quantilesort!($(rand(10^6)), false, $(rand(n)))
           @btime sort!($(rand(10^6)))
           nothing
       end
f1 (generic function with 1 method)

julia> f1(5)
  8.901 ms (0 allocations: 0 bytes)
  6.912 ms (2 allocations: 96 bytes)
  2.568 ms (0 allocations: 0 bytes)

julia> f1(50)
  18.857 ms (0 allocations: 0 bytes)
  49.713 ms (2 allocations: 480 bytes)
  1.840 ms (0 allocations: 0 bytes)

julia> f1(500)
  19.029 ms (0 allocations: 0 bytes)
  520.530 ms (8 allocations: 9.09 KiB)
  2.607 ms (0 allocations: 0 bytes)

julia> function f2(n)
           @btime Statistics._quantilesort!($(big.(rand(10^6))), false, extrema($(rand(n)))...)
           @btime new_quantilesort!($(big.(rand(10^6))), false, $(rand(n)))
           @btime sort!($(big.(rand(10^6))))
           nothing
       end
f2 (generic function with 1 method)

julia> f2(5)

  538.816 ms (0 allocations: 0 bytes)
  351.436 ms (2 allocations: 96 bytes)
  77.425 ms (0 allocations: 0 bytes)

julia> f2(50)
  861.641 ms (0 allocations: 0 bytes)
  8.368 s (2 allocations: 480 bytes)
  68.657 ms (0 allocations: 0 bytes)

julia> f2(500)
  776.346 ms (0 allocations: 0 bytes)
  51.798 s (8 allocations: 9.09 KiB)
  84.172 ms (0 allocations: 0 bytes)

```

Issue #71, opened 2021-01-24
mean! may return incorrect results


yurivish:
I noticed that the `mean!` function does not properly consider that its arguments may alias each other. In some such cases the result from `mean!` is not correct:

```julia
julia> let a = [1 2 3]
           mean!(a, a)
       end
1×3 Array{Int64,2}:
 0  0  0

julia> let a = [1 2 3]
           mean!(copy(a), a)
       end
1×3 Array{Int64,2}:
 1  2  3

julia> versioninfo()
Julia Version 1.5.3
Commit 788b2c77c1 (2020-11-09 13:37 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin18.7.0)
  CPU: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, skylake)
```

josemanuel22:
Viewing the definition of [mean!](https://docs.julialang.org/en/v1/stdlib/Statistics/#Statistics.mean!) 
The result of the operation `mean!(r, v)` is written in `r`. Therefore it seems logical to me that if you use the same object as variable `r` and as variable `v`, the result is unpredictable.

I have seen that this also happens with the `sum!` function

nalimilan:
Could you file the equivalent PR for `sum` against Julia Base? Then once a decision is made there we can apply the same change here. See https://github.com/JuliaLang/julia/issues/39385.

andreasnoack:
Closed by https://github.com/JuliaStats/Statistics.jl/pull/96

PR #96, opened 2021-11-14
adding docu to mean! explain target should not alias with the source

josemanuel22:
fix issue #71


src/Statistics.jl
```diff
@@ -107,6 +107,8 @@ mean(f, A::AbstractArray; dims=:) = _mean(f, A, dims)
     mean!(r, v)
 
 Compute the mean of `v` over the singleton dimensions of `r`, and write results to `r`.
+Note that since the mean! function is intended to operate without making any allocations,
+the target should not alias with the source.
```
andreasnoack, line 111: I think it's fine not to go into explaining why this is the case and just state
```suggestion
Note that the target must not alias with the source.
```

andreasnoack (approved):

PR #100, opened 2021-12-29
[Breaking] Update type signature to reflect documentation

### LINKED ISSUE #99: Bad error message on mean(::Int, ::Int)
```
julia> using Statistics
julia> mean(1,2)
Stacktrace:
   @ Statistics /Applications/Julia-1.7.1.app/Contents/Resources/julia/share/julia/stdlib/v1.7/Statistics/src/Statistics.jl:69
 [2] top-level scope
   @ REPL[4]:1
```

LilithHafner:
Fixes #99

### DIFF
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -58,7 +58,7 @@ julia> mean([√1, √2, √3])
 1.3820881233139908
 ```
 """
-function mean(f, itr)
+function mean(f::Function, itr)
     y = iterate(itr)
     if y === nothing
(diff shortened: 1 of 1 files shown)


LilithHafner:
Is this breaking?

nalimilan:
Thanks, but that would indeed be breaking as it won't accept functors which are not subtypes of `Function`.

LilithHafner:
In addition to breaking, this would unnecessarily limit the power of `mean`. Even if we were not concerned about backward compatibility, this would be a bad idea.

PR #104, opened 2022-02-03
Fix type instability in `var`

ParadaCarleton:


nalimilan:
Why did you close this?

PR #128, opened 2022-09-26
Prepare standalone package, step 2

nalimilan:
`mean` and `mean!` are now defined in Base.

Tests will fail until https://github.com/JuliaLang/julia/pull/46501 is merged.

### DIFF
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -11,10 +11,183 @@ using LinearAlgebra, SparseArrays
 
 using Base: has_offset_axes, require_one_based_indexing
 
-using Base: mean!, mean
-
 export cor, cov, std, stdm, var, varm, mean!, mean,
     median!, median, middle, quantile!, quantile
+
+##### mean #####
+
+if !isdefined(Base, :mean)
+    """
+        mean(itr)
+
+    Compute the mean of all elements in a collection.
+
+    !!! note
+        If `itr` contains `NaN` or [`missing`](@ref) values, the result is also
+        `NaN` or `missing` (`missing` takes precedence if array contains both).
+        Use the [`skipmissing`](@ref) function to omit `missing` entries and compute the
+        mean of non-missing values.
+
+    # Examples
+    ```jldoctest
+    julia> using Statistics
+
+    julia> mean(1:20)
+    10.5
+
+    julia> mean([1, missing, 3])
+    missing
+
+    julia> mean(skipmissing([1, missing, 3]))
+    2.0
+    ```
+    """
+    mean(itr) = mean(identity, itr)
+
+    """
+        mean(f, itr)
+
+    Apply the function `f` to each element of collection `itr` and take the mean.
+
+    ```jldoctest
+    julia> using Statistics
+
+    julia> mean(√, [1, 2, 3])
+    1.3820881233139908
+
+    julia> mean([√1, √2, √3])
+    1.3820881233139908
+    ```
+    """
+    function mean(f, itr)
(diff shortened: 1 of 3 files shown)


src/Statistics.jl
```diff
@@ -11,184 +11,11 @@ using LinearAlgebra, SparseArrays
 
 using Base: has_offset_axes, require_one_based_indexing
 
+using Base: mean!, mean
```
ararslan, line 14: Why is this necessary?

ararslan: We could do something like
```julia
if !isdefined(Base, :mean)
    export mean, mean!
end
```
which would make the intent a bit clearer, IMO, and could eventually be removed once the package requires a version of Julia that includes your `mean` PR.

nalimilan: Actually I assumed that we would merge this PR only after https://github.com/JuliaLang/julia/pull/46501 so that we don't need to support older versions. Anyway the package will only be usable on Julia 1.9 and above AFAICT.

Note that making the export conditional isn't enough to support older versions, we would also have to keep the method definitions.

nalimilan: Done.

nalimilan:
Given that we have a chicken-and-egg problem to get CI to pass in Statistics.jl and in Julia, I've restored definitions for `mean` and `mean!`, making them conditional on `isdefined(Base, :mean)`. That way we can merge this here with CI passing, and then update https://github.com/JuliaLang/julia/pull/46501 to use the latest Statistics.jl and CI should pass there too. Then we can drop function definitions. OK?

nalimilan:
I also dropped `mean` and `mean!` from the manual as that causes the Julia build to fail due to duplicate inclusion of docstrings.

ararslan:
> `mean` and `mean!` are now defined in Base.

Big if true 😜

> I also dropped `mean` and `mean!` from the manual as that causes the Julia build to fail due to duplicate inclusion of docstrings.

Documenter complains about having the same docstring twice? I guess that makes sense since in most cases that's probably user error. Unfortunate that they need to be removed though because of that though. So in the period between the merging of this and merging the linked Julia PR, `mean`(`!`) won't appear in any manual builds?

src/Statistics.jl
```diff
@@ -16,189 +16,189 @@ export cor, cov, std, stdm, var, varm, mean!, mean,
...
+
+    julia> mean([√1, √2, √3])
+    1.3820881233139908
+    ```
+    """
+    function mean(f, itr)
+        y = iterate(itr)
+        if y === nothing
+            return Base.mapreduce_empty_iter(f, +, itr,
+                                            Base.IteratorEltype(itr)) / 0
```
ararslan, line 66:
```suggestion
                                             Base.IteratorEltype(itr)) / 0
```

nalimilan:
> Documenter complains about having the same docstring twice? I guess that makes sense since in most cases that's probably user error. Unfortunate that they need to be removed though because of that though. So in the period between the merging of this and merging the linked Julia PR, `mean`(`!`) won't appear in any manual builds?

No, because the master branch of Statistics will only be used after merging the Julia PR. Any changes made here will have zero effect on users as long as we don't bump the commit in Julia.

ararslan (approved):

nalimilan:
Actually given that Statistics will remain an stdlib (at least for now), and simply become upgradable, `mean` isn't going to move to Base so this PR isn't needed anymore. https://github.com/JuliaStats/Statistics.jl/pull/148 reverts it.

PR #131, opened 2022-11-25
Fix hint on mean(::Number, ::Number)

### LINKED ISSUE #99: Bad error message on mean(::Int, ::Int)

LilithHafner:
Back before I know about `middle`, I tried `mean(1, 2)` and was briefly confused before settling on `mean((1, 2))` or `mean([1, 2])`. A better hint in the error message would have helped me.

Before:
```julia
julia> mean(1, 2)
Stacktrace:
[...]
```
After:
```julia
julia> mean(1, 2)
ERROR: ArgumentError: mean(f, itr) requires a function and an iterable.
       Perhaps you meant middle(x, y)?
Stacktrace:
[...]
```

Fixes #99

### DIFF
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -103,12 +103,12 @@ julia> mean(√, [1 2 3; 4 5 6], dims=2)
 """
 mean(f, A::AbstractArray; dims=:) = _mean(f, A, dims)
 
-function Statistics.mean(f::Number, itr::Number)
+function mean(f::Number, itr::Number)
     f_value = try
         f(itr)
     catch MethodError
-        rethrow(ArgumentError("mean(f, itr) requires a function and an iterable.
-       Perhaps you meant middle(x, y)?"))
+        rethrow(ArgumentError("""mean(f, itr) requires a function and an iterable.
+                                 Perhaps you meant middle(x, y)?""",))
     end
     Base.reduce_first(+, f_value)/1
 end
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -103,6 +103,16 @@ julia> mean(√, [1 2 3; 4 5 6], dims=2)
 """
 mean(f, A::AbstractArray; dims=:) = _mean(f, A, dims)
 
+function Statistics.mean(f::Number, itr::Number)
+    f_value = try
+        f(itr)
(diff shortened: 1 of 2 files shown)


mschauer (approved):

LilithHafner:
This adds more code just for an error message, so I can see why folks might be hesitant to merge it and/or decide to decline this PR. Is that what is happening, or has this simply fallen by the wayside?

LilithHafner:
Bump

src/Statistics.jl
```diff
@@ -103,6 +103,16 @@ julia> mean(√, [1 2 3; 4 5 6], dims=2)
 """
 mean(f, A::AbstractArray; dims=:) = _mean(f, A, dims)
 
+function mean(f::Number, itr::Number)
+    f_value = try
+        f(itr)
+    catch MethodError
+        rethrow(ArgumentError("""mean(f, itr) requires a function and an iterable.
+                                 Perhaps you meant middle(x, y)?""",))
```
nalimilan, line 111: How about not putting a newline in the error message?
```suggestion
        rethrow(ArgumentError("mean(f, itr) requires a function and an iterable. " *
                              "Perhaps you meant middle(x, y)?"))
```

nalimilan (commented):
Sorry, it's just that nobody took care of it. Sounds reasonable given that it's been reported at https://github.com/JuliaStats/Statistics.jl/pull/143 too.

BTW, a similar problem exists for `sum` in Base.

src/Statistics.jl
LilithHafner: I was following the pattern of
```
julia> (4+1)(5)
Stacktrace:
 [1] top-level scope
   @ REPL[9421]:1
```
Putting the hint on a different line as the error, but either way works.

nalimilan: Ah OK. I'm not sure whether there are strong conventions in that regard.

PR #139, opened 2023-01-20
documentation fix for quantile

mcreel:
Having read the source code and the referenced article, I believe that this PR is a more clear description of what the code does.


nalimilan (commented):
Thanks! I'm going to apply this with some changes.

PR #142, opened 2023-04-14
fix tests for error types

vtjnash:
We do not care about the actual error code here, but as seen on some of the other tests that needed similar fixes, these obstruct improvements to Base error messages.


vtjnash:
bump

vtjnash:
bump

vtjnash:
From all over the place. Sometimes implicitly in `convert`, or sometimes in `one` or `zero`

PR #143, opened 2023-04-20
Restrict type of f in `mean(f, itr)`

KronosTheLate:
A naive user (like me) might expect there to be a vararg method for `mean`, which computes the mean of all arguments if they are numbers [1]. However, if one tries to do that, a not-so-helpful error is produced:
```
julia> mean(1, 1)
Stacktrace:
   @ Statistics ~/.julia/juliaup/julia-1.9.0-rc2+0.x64.linux.gnu/share/julia/stdlib/v1.9/Statistics/src/Statistics.jl:69
 [2] top-level scope
   @ REPL[9]:1
```

This PR restricts the type of `f`, so that an appropriate `MethodError` should be thrown (I have not tested).

[1] An alternative fix would be to define such a method, as in `mean(args::Number...) = mean([args...])`. I am not sure what is preferable, but the suggested PR seems more clear.

### DIFF
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -58,7 +58,7 @@ julia> mean([√1, √2, √3])
 1.3820881233139908
 ```
 """
-function mean(f, itr)
+function mean(f::Function, itr)
     y = iterate(itr)
     if y === nothing
         return Base.mapreduce_empty_iter(f, +, itr,


nalimilan:
Unfortunately we can't do that, as we need to allow any callable object, which includes any type. The same problem applies to many methods in Base.

nalimilan:
Also note that `sum(1, 1)` has the same problem. This should be changed in Base first if we wanted to change this here.

nalimilan:
I don't think this can be fixed, so you could try to discuss this in Base but I'm pretty sure about the outcome. But see https://github.com/JuliaStats/Statistics.jl/pull/131 for an alternative "fix".

KronosTheLate:
Alright, that seems like a better fix - to dispatch only on the case with numbers passed. So then I know that this issue is still noted and worked on somewhere - thanks ^_^

LilithHafner:
Duplicate of #100

PR #145, opened 2023-07-01
Fix overflows in `quantile`

nalimilan:
The `a + γ*(b-a)` introduced by https://github.com/JuliaLang/julia/pull/16572 has the advantage that it increases with `γ` even when `a` and `b` are very close, but it has the drawback that it is not robust to overflow. This is likely to happen in practice with small integer and floating point types.

Conversely, the `(1-γ)*a + γ*b` which is currently used only for non-finite quantities is robust to overflow but may not always increase with `γ` as when `a` and `b` are very close or (more frequently) equal since precision loss can give a slightly smaller value for a larger `γ`. This can be problematic as it breaks an expected invariant.

So keep using the `a + γ*(b-a)` formula when `a ≈ b`, in which case it's almost like returning either `a` or `b` but less arbitrary.

Fixes https://github.com/JuliaStats/Statistics.jl/issues/144.

### DIFF
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -1020,7 +1020,10 @@ end
         b = v[j + 1]
     end
 
-    if isfinite(a) && isfinite(b)
+    # When a ≉ b, b-a may overflow
+    # When a ≈ b, (1-γ)*a + γ*b may not be increasing with γ due to rounding
+    # Call to float is to work around JuliaLang/julia#50380
+    if isfinite(a) && isfinite(b) && float(a) ≈ float(b)
         return a + γ*(b-a)
     else
         return (1-γ)*a + γ*b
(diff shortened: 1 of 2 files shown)


bkamins (approved):
It required also testing if the function is non-decreasing if we increase `b` and switch the formula, but I tested it and it holds.

nalimilan:
> It required also testing if the function is non-decreasing if we increase `b` and switch the formula, but I tested it and it holds.

It's already covered by the test added a long time ago by https://github.com/JuliaLang/julia/pull/16572. That's how I realized the problem. ;-)

bkamins:
In general I mean that it should be monotonic in `a`, `b` and `γ` and I checked all. The tests did not cover all cases fully. The reason is that e.g. you need to test case when `float(a) ≈ float(b)` but `!(float(a) ≈ nextfloat(float(b)))` (and similarly `!(prevfloat(float(a)) ≈ float(b))` for monotonicity for various `γ`; I think it is not covered but I checked it).

nalimilan:
OK. So you mean two tests like this are needed?
```julia
    @test issorted(quantile([1.0, 1.0+eps(), 1.0+2eps(), 1.0+3eps()], range(0, 1, length=100)))
    @test issorted(quantile([1.0, 1.0+2eps(), 1.0+4eps(), 1.0+6eps()], range(0, 1, length=100)))
```

bkamins:
Yes - something like this (this is not strictly needed 😄, but I run such tests and they were OK).

PR #134, opened 2022-12-07
Make SparseArrays a weak dependency

IanButterworth:
Uses the recently merged https://github.com/JuliaLang/julia/pull/47695 to set up SparseArrays as a weak dependency of Statistics, via a package extension.

All code was copied across to the package extension untouched, with the addition of some imports.

This is backwards compatible. On older julia versions that don't support package extensions, SparseArrays will be loaded as per normal.

What this would look like on julia master, if SparseArrays is removed from the sysimage
```
julia> @time using Statistics
  0.011658 seconds (13.76 k allocations: 1.030 MiB)

julia> @time using SparseArrays
  4.618848 seconds (7.05 M allocations: 980.989 MiB, 1.46% compilation time)
```

cc. @KristofferC

PR #146, opened 2023-07-04
Handle SparseArrays.jl dep as Pkg extension

dkarrasch:
I wonder if we want to/can/should backport this to v1.9 and v1.10?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,15 +3,22 @@ uuid = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 keywords = ["statistics"]
 license = "MIT"
 desc = "Basic statistics for Julia."
-version = "1.9.0"
+version = "1.10.0"
 
 [deps]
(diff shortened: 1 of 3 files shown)


dkarrasch:
Well, backporting to v1.9 doesn't make much sense, I guess, since SparseArrays.jl is in the sysimage. But backporting to v1.10 is important.

dkarrasch:
Uff, I only now realize this is a duplicate of #134. What needs to be done to get this done? Not having SparseArrays.jl as a weak dependency puts a heavy load (time) on downstream packages like Distances.jl.

nalimilan:
Not sure. If it is confirmed that weak dependencies don't work for stdlib modules, we will have to finish https://github.com/JuliaLang/julia/pull/46501 first.

dkarrasch:
Right, but the Statistics.jl dependency is not touched here. We only make the direct dependency on SparseArrays.jl weak.

PR #147, opened 2023-08-15
Recommend `mean((x, y))` rather than `middle((x, y))`

nalimilan:
It seems more logical and simpler for users to recommend using the same function. This also helps making Statistics a standalone package (JuliaStats/Statistics.jl#128) as `mean` will be moved to Julia Base, but `middle` will remain in Statistics.

This blocks https://github.com/JuliaLang/julia/pull/46501.

See https://github.com/JuliaStats/Statistics.jl/pull/131. Cc: @LilithHafner

### DIFF
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -107,9 +107,13 @@ if !isdefined(Base, :mean)
     function mean(f::Number, itr::Number)
         f_value = try
             f(itr)
-        catch MethodError
-            rethrow(ArgumentError("""mean(f, itr) requires a function and an iterable.
-                                    Perhaps you meant mean((x, y))?"""))
+        catch err
+            if err isa MethodError && err.f === f
+                rethrow(ArgumentError("""mean(f, itr) requires a function and an iterable.
+                                         Perhaps you meant mean((x, y))?"""))
+            else
+                rethrow(err)
+            end
         end
         Base.reduce_first(+, f_value)/1
     end
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -108,7 +108,7 @@ if !isdefined(Base, :mean)
         f_value = try
             f(itr)
         catch err
-            if err isa MethodError && err.f === f
+            if err isa MethodError && err.f === f && err.args == (itr,)
                 rethrow(ArgumentError("""mean(f, itr) requires a function and an iterable.
                                          Perhaps you meant mean((x, y))?"""))
             else
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -109,7 +109,7 @@ if !isdefined(Base, :mean)
             f(itr)
         catch MethodError
             rethrow(ArgumentError("""mean(f, itr) requires a function and an iterable.
-                                    Perhaps you meant middle(x, y)?""",))
+                                    Perhaps you meant mean((x, y))?"""))
         end
         Base.reduce_first(+, f_value)/1
     end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -179,7 +179,7 @@ end
     end
     (t::T)(y) = t.x == 0 ? t(y, y + 1, y + 2) : t.x * y
     @test mean(T(2), 3) === 6.0
-    err = @test_throws MethodError mean(T(0), 3)
+    @test_throws MethodError mean(T(0), 3)
     struct U <: Number
         x::Int
     end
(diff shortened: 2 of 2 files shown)


bkamins (approved):

src/Statistics.jl
```diff
@@ -109,7 +109,7 @@ if !isdefined(Base, :mean)
             f(itr)
         catch MethodError
```
ararslan, line 110: Unrelated to this PR but notably not how to catch a `MethodError`, haha

src/Statistics.jl
```diff
@@ -109,7 +109,7 @@ if !isdefined(Base, :mean)
...
             rethrow(ArgumentError("""mean(f, itr) requires a function and an iterable.
-                                    Perhaps you meant middle(x, y)?""",))
+                                    Perhaps you meant mean((x, y))?"""))
```
ararslan, line 112: Are there any edge cases where these are not equivalent?

ararslan: Here's one: https://github.com/JuliaStats/Statistics.jl/issues/140#issuecomment-1597240212

```julia
julia> mean((floatmax(Float32), floatmax(Float32)))
Inf32

julia> middle(floatmax(Float32), floatmax(Float32))
3.4028235f38
```

src/Statistics.jl
nalimilan: Woops. I've pushed a commit to make the check stricter.

src/Statistics.jl
bkamins: I agree. When user writes `mean(x, y)` then they probably expect it to be the same as `mean` of collection containing `x` and `y`.

nalimilan:
[Apparently](https://github.com/JuliaLang/julia/pull/46501#issuecomment-1679253193) this won't be required to make Statistics an upgradable stdlib, but it's probably a good idea anyway.

src/Statistics.jl
ararslan: Okay, makes sense.

ararslan (approved):

src/Statistics.jl
```diff
@@ -107,9 +107,13 @@ if !isdefined(Base, :mean)
     function mean(f::Number, itr::Number)
         f_value = try
             f(itr)
-        catch MethodError
-            rethrow(ArgumentError("""mean(f, itr) requires a function and an iterable.
-                                    Perhaps you meant middle(x, y)?""",))
+        catch err
+            if err isa MethodError && err.f === f
```
ararslan, line 111: Come to think of it, would it also be worth checking that `err.args` is `(itr,)`? Otherwise the caught exception may not correspond to a missing method for the call that tried. For (admittedly contrived) example:
```julia
julia> struct Cursed <: Number
           value::Int
       end

julia> (evil::Cursed)(x::Number) = evil(x, x + 1, x + 2)

julia> bad = Cursed(666)
Cursed(666)

julia> ex = try
           bad(69)
       catch ex
           ex
       end
MethodError(Cursed(666), (69, 70, 71), 0x0000000000007b17)

julia> ex isa MethodError && ex.f === bad
true

julia> ex.args == (69,)
false
```
But perhaps that's such a ridiculous case that it's not worth thinking about...

src/Statistics.jl
LilithHafner: > Unrelated to this PR but notably not how to catch a `MethodError`, haha

haha, that was me :P. Oops.

LilithHafner (approved):

src/Statistics.jl
nalimilan: OK, why not, I've added a commit.

test/runtests.jl
```diff
@@ -173,12 +173,18 @@ end
...
-    @test_throws "ArgumentError: mean(f, itr) requires a function and an iterable.\nPerhaps you meant middle(x, y)" mean(1, 2)
+    @test_throws "ArgumentError: mean(f, itr) requires a function and an iterable.\nPerhaps you meant mean((x, y))" mean(1, 2)
     struct T <: Number
         x::Int
     end
-    (t::T)(y) = t.x * y
-    @test @inferred mean(T(2), 3) === 6.0
+    (t::T)(y) = t.x == 0 ? t(y, y + 1, y + 2) : t.x * y
+    @test mean(T(2), 3) === 6.0
+    err = @test_throws MethodError mean(T(0), 3)
```
ararslan, line 182:
```suggestion
    @test_throws MethodError mean(T(0), 3)
```
AFAICT `err` isn't being used.

ararslan (approved):

PR #148, opened 2023-08-25
Revert "Prepare standalone package, step 2 (#128)"

nalimilan:
This reverts commit 14438ab59ec5ea76d6eea844f03040c692d3b1f3. `mean` will not be moved to Base as Statistics will remain an (upgradable) stdlib.

Keep version at 1.11.0 though.

### DIFF
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -16,188 +16,188 @@ export cor, cov, std, stdm, var, varm, mean!, mean,
 
 ##### mean #####
 
-if !isdefined(Base, :mean)
-    """
-        mean(itr)
+"""
+    mean(itr)
 
-    Compute the mean of all elements in a collection.
+Compute the mean of all elements in a collection.
 
-    !!! note
-        If `itr` contains `NaN` or [`missing`](@ref) values, the result is also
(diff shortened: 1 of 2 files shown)


src/Statistics.jl
```diff
@@ -16,193 +16,193 @@ export cor, cov, std, stdm, var, varm, mean!, mean,
...
+        return result ./= n
+    end
+end
+
+function mean(r::AbstractRange{<:Real})
+    isempty(r) && return oftype((first(r) + last(r)) / 2, NaN)
+    (first(r) + last(r)) / 2
 end
 
+median(r::AbstractRange{<:Real}) = mean(r)
```
bkamins, line 204: why is this definition  moved here?

bkamins (approved):

src/Statistics.jl
nalimilan: In the commit this reverts I had moved it outside of the `if` as we had to define it even if `mean` was defined in Base. I'm just moving it to its original place. But yeah the new one is more logical so maybe we could live it there.

bkamins: For me the location near other `median` methods is more logical.

nalimilan: OK, let's do that.

PR #149, opened 2023-08-31
move SparseArrays to an extension

vtjnash:


vtjnash:
Oops, https://github.com/JuliaStats/Statistics.jl/pull/134 is better than this

PR #150, opened 2023-09-01
Prevent overflow in `mean(::AbstractRange)` and relax type constraint.

chunjiw:
This is almost an exact copy of #115. This should fix https://github.com/JuliaLang/julia/issues/50982 as well.

My understanding is that #115 is still open because too many commits are ahead of it. So someone needs to make a similar pull request. So here it is.... I have contributed to open source before but not experienced, so please let me know if I missed anything.

### DIFF
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -198,8 +198,8 @@ if !isdefined(Base, :mean)
     end
 
     function mean(r::AbstractRange{T}) where T
-        isempty(r) && return zero(T) / 0
-        return first(r) / 2 + last(r) / 2
+        isempty(r) && return zero(T)/0
+        return first(r)/2 + last(r)/2
     end
 end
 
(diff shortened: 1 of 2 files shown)


mschauer:
Looks good!

src/Statistics.jl
```diff
@@ -197,9 +197,9 @@ if !isdefined(Base, :mean)
         end
     end
 
-    function mean(r::AbstractRange{<:Real})
-        isempty(r) && return oftype((first(r) + last(r)) / 2, NaN)
-        (first(r) + last(r)) / 2
+    function mean(r::AbstractRange{T}) where T
+        isempty(r) && return zero(T) / 0
+        return first(r) / 2 + last(r) / 2
```
nalimilan, line 202: Probably more readable this way?
```suggestion
        isempty(r) && return zero(T)/0
        return first(r)/2 + last(r)/2
```

nalimilan (approved):
Looks good to me, thanks!

ararslan (approved):

PR #151, opened 2023-09-09
Call user function only once in `mean`

### LINKED ISSUE #49: Statistic.mean(f, A) calls f length(A)+1 times
Was expecting an iterator to only call the function once per element.   

```
"""
Testing mean calls F length+1 times
"""
function meantest()
    aa = [1, 3, 2, 7, 11]
    bb = [1, 2, 3, 4, 5]


    function meanf(a)
        b = 1
        d = mean(a) do v
            r = v+b
            println("$b $r $v")
            b += 1
            return r
        end
        return d
    end

    avg = sum(aa .+ bb) / length(aa)
    avg2 = meanf(aa)
    println(" $avg != $avg2 ")
end


```

julia> versioninfo()
Julia Version 1.5.0-rc2.0
Commit 7f0ee122d7 (2020-07-27 15:24 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin19.5.0)
  CPU: Intel(R) Core(TM) i7-3615QM CPU @ 2.30GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, ivybridge)
Environment:
  JULIA_NUM_THREADS = 8

nalimilan:
Use an internal type to initiate promotion instead of using the type returned by user-provided function applied to the first element. This avoid calling the function twice, which can be confusing.

Fixes #49 but with a more general solution than https://github.com/JuliaStats/Statistics.jl/pull/80/.

Cc: @stevengj @kagalenko-m-b 

This doesn't make any difference for performance:
```julia
julia> x = rand(100_000_000);

julia> y = ifelse.(rand(Bool, length(y)) .> .25, missing, x);

julia> z = Vector{Union{Missing, Float64, Int}}(y);

julia> z .= ifelse.(rand(Bool, length(y)) .> .25, round.(Int, x), z);

# master
julia> @btime mean(x);
  68.582 ms (1 allocation: 16 bytes)

julia> @btime mean(y);
  472.870 ms (1 allocation: 16 bytes)

julia> @btime mean(z);
  533.646 ms (3 allocations: 48 bytes)

# PR
julia> @btime mean(x);
  69.386 ms (1 allocation: 16 bytes)

julia> @btime mean(y);
  476.337 ms (0 allocations: 0 bytes)

julia> @btime mean(z);
  557.560 ms (0 allocations: 0 bytes)
```


nalimilan:
The new commit implements a simpler approach which is fast for concrete eltypes but super slow for complex type unions as the compiler doesn't infer an element type more precise than `Any`
```julia
julia> @btime mean(x);
  68.529 ms (4 allocations: 64 bytes)

julia> @btime mean(y);
  550.489 ms (0 allocations: 0 bytes)

julia> @btime mean(z);
  2.752 s (25300984 allocations: 386.06 MiB)

kagalenko-m-b:
It looks like you're reimplementing undocumented `Base` internals, doesn't that have potential to break in future Julia versions?

nalimilan:
Yes, but given that Statistics is an stdlib any breakage would be noticed in Julia even before merging it. I'm more concerned about performance issues introduced by the PR...

PR #152, opened 2023-09-11
Document MATLAB behavior in `quantile` docstring

nalimilan:
This recently [raised a question](https://stackoverflow.com/questions/77077459/quantile-function-julia-vs-matlab) on StackOverflow.


mschauer (approved):

bkamins (approved):
So it looks that someone reads these SO questions and answers :).

PR #157, opened 2023-11-03
Drop support for v1.9 in CI

vtjnash:
The package was already broken by https://github.com/JuliaStats/Statistics.jl/pull/153, so this merely updates the CI support matrix and compat bounds so that someone doesn't accidentally install a newer version than is supported by v1.9

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
       matrix:
         version:
           #- '1' # automatically expands to the latest stable 1.x release of Julia
-          - '1.10'
+          #- '1.10'
           - 'nightly'
         os:
           - ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -12,7 +12,8 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1' # automatically expands to the latest stable 1.x release of Julia
+          #- '1' # automatically expands to the latest stable 1.x release of Julia
+          #- '1.10'
```
giordano, line 16: If you want to test on 1.10.0-rc1:
```suggestion
          - '~1.10.0-0'
```

vtjnash: I mean, I hate that syntax, but I do want that CI more

nalimilan:
Thanks!

PR #169, opened 2024-06-26
correct std docs: uses abs2 & sqrt is elementwise

stevengj:
The docstring for `std` was missing a `.` to indicate that the `sqrt` is computed elementwise.  Also, it uses `abs2` rather than `^2` (which is different for complex-valued data).

cc @lxvm

### DIFF
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -441,7 +441,7 @@ The algorithm returns an estimator of the generative distribution's standard
 deviation under the assumption that each entry of `itr` is a sample drawn from
 the same unknown distribution, with the samples uncorrelated.
 For arrays, this computation is equivalent to calculating
-`sqrt(sum((itr .- mean(itr)).^2) / (length(itr) - 1))`.
+`sqrt.(sum((itr .- mean(itr)).^2) / (length(itr) - 1))`.
 If `corrected` is `true`, then the sum is scaled with `n-1`,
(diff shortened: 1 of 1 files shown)


stevengj:
bump.  @andreasnoack?

src/Statistics.jl
```diff
@@ -441,7 +441,7 @@ The algorithm returns an estimator of the generative distribution's standard
 deviation under the assumption that each entry of `itr` is a sample drawn from
 the same unknown distribution, with the samples uncorrelated.
 For arrays, this computation is equivalent to calculating
-`sqrt(sum((itr .- mean(itr)).^2) / (length(itr) - 1))`.
+`sqrt.(sum(abs2.(itr .- mean(itr))) / (length(itr) - 1))`.
```
andreasnoack, line 444: Unless `dims` is specified then the output from `sum` is a scalar, no?

stevengj: Maybe I was thinking of the case with a dims argument. (Though the source uses `sqrt.` in cases without dims too.)

@lxvm, do you remember the case we came across that seemed ambiguous without the dot? It’s been a couple weeks and I don’t remember why this came up.

andreasnoack: Ah yes, in that case the function works element wise

andreasnoack (approved):

PR #178, opened 2024-10-20
Test earliest supported Julia version and prereleases

devmotion:
#167 removed tests on nightly and added the `include-prereleases` option. However, I think it would be better to test `pre` instead of setting this option: Otherwise, once a pre-release of Julia v1 is available, tests with `1` will only test the pre-release but the latest stable build will not be tested anymore.

Additionally, I think it would be good to test `min`, the earliest supported Julia version, as well to avoid accidentally breaking support for old Julia versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,31 +6,26 @@ on:
   pull_request:
 jobs:
   test:
-    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
+    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ github.event_name }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         version:
-          - '1' # automatically expands to the latest stable 1.x release of Julia
-          - 'lts'
+          - 'min' # earliest supported version of Julia
+          - '1' # latest v1 version of Julia
+          - 'lts' # latest LTS build of Julia
+          - 'pre' # latest prerelease build (RCs, betas, and alphas) of Julia
         os:
           - ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -6,31 +6,26 @@ on:
...
-    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
+    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ github.event_name }}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         version:
-          - '1' # automatically expands to the latest stable 1.x release of Julia
-          - 'lts'
+          - 'min' # earliest supported version of Julia
```
andreasnoack, line 15: I'm wondering if `min` is necessary given that this is a stdlib

ViralBShah:
If you don't mention arch, does it just pick aarch64 for mac and x64 for all the others? Just should make sure that we are not doing x64 in emulation on macOS-latest, and we are indeed running natively.

.github/workflows/ci.yml
devmotion: Hmm maybe not. I added it since Statistics specifies a `julia = 1.9.4` compat.

devmotion:
> If you don't mention arch, does it just pick aarch64 for mac and x64 for all the others?

Yes, by default the setup-julia action uses the Julia binaries for the architecture of the runner, ie aarch64 on macOS-latest. This can be seen in e.g. https://github.com/JuliaStats/Statistics.jl/actions/runs/11428315624/job/31793421540?pr=178#step:3:8

ViralBShah:
Hmm, I might have hit merge a bit early.

PR #184, opened 2025-01-02
Use muladd in aleph calculation in _quantile to avoid some rounding errors

andreasnoack:
In some cases when aleph is an integer (in exact) arithmetic, the calculation of aleph via m can introduce some rounding errors that makes the value non-integer. Using muladd avoids this in some cases.

### DIFF
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -1035,8 +1035,10 @@ end
     @assert n > 0 # this case should never happen here
 
     m = alpha + p * (one(alpha) - alpha - beta)
-    aleph = n*p + oftype(p, m)
-    j = clamp(trunc(Int, aleph), 1, n-1)
+    # Using muladd here avoids some rounding errors when aleph is an integer
+    # The use of oftype supresses the promotion caused by alpha and beta
+    aleph = muladd(n, p, oftype(p, m))
+    j = clamp(trunc(Int, aleph), 1, n - 1)
     γ = clamp(aleph - j, 0, 1)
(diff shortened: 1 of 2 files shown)


src/Statistics.jl
```diff
@@ -1035,8 +1035,10 @@ end
     @assert n > 0 # this case should never happen here
 
     m = alpha + p * (one(alpha) - alpha - beta)
-    aleph = n*p + oftype(p, m)
-    j = clamp(trunc(Int, aleph), 1, n-1)
+    # Using muladd here avoids some rounding errors when aleph is an integer
+    # The use of oftype supresses the promotion caused by alpha and beta
+    aleph = muladd(n, p, oftype(p, m))
```
nalimilan, line 1040: AFAIK `muladd` only aims at performance, it's `fma` which aims at precision (even if in practice both are often the same I think). Performance isn't an issue here so we're fine.
```suggestion
    # Using fma here avoids some rounding errors when aleph is an integer
    # The use of oftype supresses the promotion caused by alpha and beta
    aleph = fma(n, p, oftype(p, m))
```

nalimilan (approved):

KristofferC:
My guess is that this broke the doctests on some julia CI workers (https://github.com/JuliaLang/julia/pull/57238#issuecomment-2630242407). I haven't verified.

nalimilan:
Indeed. See https://github.com/JuliaStats/Statistics.jl/pull/188.

PR #187, opened 2025-01-22
Support non-interpolating quantile definitions

### LINKED ISSUE #185: Support for non-interpolating quantile computation
In the taxonomy of https://www.tandfonline.com/doi/abs/10.1080/00031305.1996.10473566 (see summary on [Wikipedia](https://en.wikipedia.org/wiki/Quantile#Estimating_quantiles_from_a_sample)), this would be the R-1 definition and is identical to the "population quantile" defined on Wikipedia. We currently only support the 4-9 definitions through the use of the `alpha` and `beta` keyword arguments.

The motivation is that it is sometimes nice that quantiles are based on values in the input data. This is something we have previously discussed in the context of the `median` function and it recently came up in https://github.com/JuliaData/CategoricalArrays.jl/issues/381#issuecomment-2568004308.

Since the R-1 definition can't be reached through special values of `alpha` and `beta` we'd need a separate interface if we are to support the new definition. One possibility is to introduce a new keyword argument, say `method` or `definition`, which takes `Enum` values and errors out if set simultaneously with `alpha`/`beta`.

nalimilan:
Add a `type` argument to `quantile` to support the three remaining types that we didn't support. Some of these are useful in particular because they correspond to actual values from the data and work for types that do not support arithmetic.

Fixes #185.

### DIFF
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -956,7 +956,7 @@ Definitions 4 to 9 use linear interpolation between consecutive order statistics
 Samples quantiles are defined by `Q(p) = (1-γ)*x[j] + γ*x[j+1]`,
 where `j = floor(n*p + m)`, `m = alpha + p*(1 - alpha - beta)` and `γ = n*p + m - j`.
 - `type=4`: `alpha=0`, `beta=1` (SAS-1)
-- `type=5`: `alpha=0.5`, `beta=0.5` (MATLAB default)
+- `type=5`: `alpha=1/2`, `beta=1/2` (MATLAB default)
 - `type=6`: `alpha=0`, `beta=0` (Excel `PERCENTILE.EXC`, Python default, Stata `altdef`)
 - `type=7`: `alpha=1`, `beta=1` (Julia, R and NumPy default, Excel `PERCENTILE` and
   `PERCENTILE.INC`, Python `'inclusive'`)
@@ -1080,16 +1080,16 @@ function _quantile(v::AbstractVector, p::Real, ::Val{type},
         type === nothing ||
             throw(ArgumentError("it is not allowed to pass both `type` and `alpha` or `beta`"))
 
-        alpha === nothing && (alpha = 1.0)
+        alpha === nothing && (alpha = 1)
         beta === nothing && (beta = alpha)
 
         0 <= alpha <= 1 || throw(ArgumentError("alpha parameter out of [0,1] range"))
         0 <= beta <= 1 || throw(ArgumentError("beta parameter out of [0,1] range"))
     elseif type === nothing
-        alpha = beta = 1.0
+        alpha = beta = 1
     elseif 4 <= type <= 9
-        alpha = (0.0, 1/2, 0.0, 1.0, 1/3, 3/8)[type-3]
-        beta  = (1.0, 1/2, 0.0, 1.0, 1/3, 3/8)[type-3]
+        alpha = oftype(p, (0, 1//2, 0, 1, 1//3, 3//8)[type-3])
+        beta  = oftype(p, (1, 1//2, 0, 1, 1//3, 3//8)[type-3])
     elseif !(1 <= type <= 3)
         throw(ArgumentError("`type` must be between 1 and 9"))
     end
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -854,8 +854,12 @@ median!(v::AbstractArray) = median!(vec(v))
     median(itr)
 
 Compute the median of all elements in a collection.
-For an even number of elements no exact median element exists, so the result is
-equivalent to calculating mean of two median elements.
+
+For an even number of elements no exact median element exists, so the
+mean of two median elements is returned.
+This is equivalent to [`quantile(itr, 0.5, type=2)`](@ref).
+Use `quantile` with `type=1` or `type=3` to compute median of types
+with limited or no support for arithmetic operations, such as `Date`.
 
 !!! note
     If `itr` contains `NaN` or [`missing`](@ref) values, the result is also
@@ -923,7 +927,8 @@ julia> median([√1, √3, √2])
 median(f::Function, v) = median!(f.(v))
 
 """
-    quantile!([q::AbstractArray, ] v::AbstractVector, p; sorted=false, alpha::Real=1.0, beta::Real=alpha)
+    quantile!([q::AbstractArray, ] v::AbstractVector, p;
+              sorted=false, type::Integer=7, alpha::Real=1.0, beta::Real=alpha)
 
 Compute the quantile(s) of a vector `v` at a specified probability or vector or tuple of
 probabilities `p` on the interval [0,1]. If `p` is a vector, an optional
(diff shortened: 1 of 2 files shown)


src/Statistics.jl
```diff
@@ -1024,65 +1048,112 @@ function _quantilesort!(v::AbstractVector, sorted::Bool, minp::Real, maxp::Real)
...
-- Def. 9: `alpha=3/8`, `beta=3/8`
+The keyword argument `type` can be used to choose among the 9 definitions
+in Hyndman and Fan (1996). Alternatively, `alpha` and `beta` allow reproducing
+any of the methods 4-9 defined in this paper. It is not allowed to specify both
+kinds of arguments at the same time.
+
+Definitions 1 to 3 are discontinuous:
+- `type=1`: `Q(p) = x[ceil(n*p)]` (SAS-3)
+- `type=2`: `Q(p) = middle(x[ceil(n*p), floor(n*p + 1)])` (SAS-5, Stata)
+- `type=3`: `Q(p) = x[round(n*p)]` (SAS-2)
```
nalimilan, line 1142: Here I have even simplified formulas a bit more and don't even mention clamping at the extremes. Maybe that's too much and it should at least be mentioned somewhere (e.g. in the sentence common to the three types)?

test/runtests.jl
```diff
@@ -797,6 +806,46 @@ end
     @test quantile(v, 1.0, alpha=0.0, beta=0.0) ≈ 21.0
     @test quantile(v, 1.0, alpha=1.0, beta=1.0) ≈ 21.0
 
+    # tests against R's quantile with type=1
+    @test quantile(v, 0.0, type=1) === 2
```
nalimilan, line 810: As you can see the type of the result can be different for types 1 and 3 as we keep the original type instead of using whatever type arithmetic operations produce. This makes sense and is even necessary if we want to work for types that don't support arithmetic.

But this means the inferred return type when passing `type` is a `Union` of two types. Maybe OK as it's small enough to be optimized out? If not there are probably ways to ensure inference works via combined use of `Val(type)` and `@inline` in a wrapper function.

(Note that when omitting `type`, the inferred type is concrete as before so at least there's no regression.)

src/Statistics.jl
```diff
@@ -905,31 +909,44 @@ _median(v::AbstractArray{T}, ::Colon) where {T} = median!(copyto!(Array{T,1}(und
...
-- Def. 8: `alpha=1/3`, `beta=1/3`
-- Def. 9: `alpha=3/8`, `beta=3/8`
+The keyword argument `type` can be used to choose among the 9 definitions
+in Hyndman and Fan (1996). Alternatively, `alpha` and `beta` allow reproducing
+any of the methods 4-9 defined in this paper. It is not allowed to specify both
+kinds of arguments at the same time.
+
+Definitions 1 to 3 are discontinuous:
+- `type=1`: `Q(p) = x[ceil(n*p)]` (SAS-3)
+- `type=2`: `Q(p) = middle(x[ceil(n*p), floor(n*p + 1)])` (SAS-5, Stata)
```
bkamins, line 934: Notation, you meant: `x[[ceil(n*p), floor(n*p + 1)]]` instead?
Also `floor(n*p + 1)` can be out of bounds (for `p=1`)

src/Statistics.jl
```diff
@@ -905,31 +909,44 @@ _median(v::AbstractArray{T}, ::Colon) where {T} = median!(copyto!(Array{T,1}(und
...
 The keyword argument `sorted` indicates whether `v` can be assumed to be sorted; if
 `false` (the default), then the elements of `v` will be partially sorted in-place.
 
-Samples quantile are defined by `Q(p) = (1-γ)*x[j] + γ*x[j+1]`,
-where `x[j]` is the j-th order statistic of `v`, `j = floor(n*p + m)`,
-`m = alpha + p*(1 - alpha - beta)` and `γ = n*p + m - j`.
-
-By default (`alpha = beta = 1`), quantiles are computed via linear interpolation between the points
-`((k-1)/(n-1), x[k])`, for `k = 1:n` where `n = length(v)`. This corresponds to Definition 7
+By default (`type=7`, or equivalently `alpha = beta = 1`),
```
bkamins, line 921: why do we use `x` and `itr`, while the function accepts `v` as its argument?

src/Statistics.jl
```diff
@@ -905,31 +909,44 @@ _median(v::AbstractArray{T}, ::Colon) where {T} = median!(copyto!(Array{T,1}(und
...
+any of the methods 4-9 defined in this paper. It is not allowed to specify both
+kinds of arguments at the same time.
+
+Definitions 1 to 3 are discontinuous:
+- `type=1`: `Q(p) = x[ceil(n*p)]` (SAS-3)
+- `type=2`: `Q(p) = middle(x[ceil(n*p), floor(n*p + 1)])` (SAS-5, Stata)
+- `type=3`: `Q(p) = x[round(n*p)]` (SAS-2)
+
+Definitions 4 to 9 use linear interpolation between consecutive order statistics.
+Samples quantiles are defined by `Q(p) = (1-γ)*x[j] + γ*x[j+1]`,
```
bkamins, line 938: again `x[j+1]` can be out of bounds in some cases (e.g. for `p=1`)

src/Statistics.jl
```diff
@@ -1024,65 +1048,112 @@ function _quantilesort!(v::AbstractVector, sorted::Bool, minp::Real, maxp::Real)
...
+        j = clamp(floor(Int, n*p + 1), 1, n)
+        return middle(v[i], v[j])
+    elseif type == 3
+        return v[clamp(round(Int, n*p), 1, n)]
     else
-        a = v[j]
-        b = v[j + 1]
-    end
+        m = alpha + p * (one(alpha) - alpha - beta)
+        # Using fma here avoids some rounding errors when aleph is an integer
```
bkamins, line 1089:
```suggestion
        # Using fma here avoids some rounding errors when alpha is an integer
```

src/Statistics.jl
```diff
@@ -1024,65 +1048,112 @@ function _quantilesort!(v::AbstractVector, sorted::Bool, minp::Real, maxp::Real)
...
+        # The use of oftype supresses the promotion caused by alpha and beta
+        aleph = fma(n, p, oftype(p, m))
```
bkamins, line 1091: this will error if `p=0` or `p=1` and `m` is a fraction.

src/Statistics.jl
```diff
@@ -1024,65 +1048,112 @@ function _quantilesort!(v::AbstractVector, sorted::Bool, minp::Real, maxp::Real)
...
+        j = clamp(trunc(Int, aleph), 1, n - 1)
```
bkamins, line 1092: is this clamp correct if `p=1`? It seems to me that then `j` should be `n` and later we just need to make sure not to use `j+1`.

src/Statistics.jl
bkamins: This was my thought I left above.

test/runtests.jl
bkamins: I would add tests for `p=0` and `p=1` (i.e. integer `p`), or even `p=false` and `p=true` (I undesrand we allow it).

src/Statistics.jl
nalimilan: Good catch. Though this PR doesn't touch this code, let's handle it separately?

Reproducer:
```julia
quantile(1:3, 0, alpha=0.2, beta=0.0)
```

src/Statistics.jl
nalimilan: I think so, because `γ` is 1 in that case and we return `(1-γ)*v[j] + γ*v[j + 1]` (so we don't care about `v[j]`).

src/Statistics.jl
nalimilan: I've added a sentence, let me know what you think.

nalimilan:
Gentle bump. :-)

test/runtests.jl
andreasnoack: Shouldn't constant propagation be able to handle this? I.e. make `quantile_type1(v, p) = quantile(v, q, type=1)` inferred?

src/Statistics.jl
```diff
@@ -1042,76 +1069,127 @@ function _quantilesort!(v::AbstractVector, sorted::Bool, minp::Real, maxp::Real)
...
+        alpha === nothing && (alpha = 1.0)
+        beta === nothing && (beta = alpha)
+
+        0 <= alpha <= 1 || throw(ArgumentError("alpha parameter out of [0,1] range"))
+        0 <= beta <= 1 || throw(ArgumentError("beta parameter out of [0,1] range"))
+    elseif type === nothing
+        alpha = beta = 1.0
+    elseif 4 <= type <= 9
+        alpha = (0.0, 1/2, 0.0, 1.0, 1/3, 3/8)[type-3]
+        beta  = (1.0, 1/2, 0.0, 1.0, 1/3, 3/8)[type-3]
```
andreasnoack, line 1091: I know it is no different from the current implementation but are we certain that these values should be `Float64`? E.g. if all arguments were `Float32` and you wanted to avoid intermediate promotion. For elements with higher precision, I suppose this would also mean that there'd be a loss of precision because `1/3` is not representable as a binary float. Since it only affects `type=8`, and therefore a fairly negligible corner case, we can maybe just open an issue to have the loss on record instead of dealing with it here.

src/Statistics.jl
```diff
@@ -923,31 +927,47 @@ julia> median([√1, √3, √2])
...
-The keyword arguments `alpha` and `beta` correspond to the same parameters in Hyndman and Fan,
-setting them to different values allows to calculate quantiles with any of the methods 4-9
-defined in this paper:
-- Def. 4: `alpha=0`, `beta=1`
-- Def. 5: `alpha=0.5`, `beta=0.5` (MATLAB default)
-- Def. 6: `alpha=0`, `beta=0` (Excel `PERCENTILE.EXC`, Python default, Stata `altdef`)
-- Def. 7: `alpha=1`, `beta=1` (Julia, R and NumPy default, Excel `PERCENTILE` and `PERCENTILE.INC`, Python `'inclusive'`)
-- Def. 8: `alpha=1/3`, `beta=1/3`
-- Def. 9: `alpha=3/8`, `beta=3/8`
+The keyword argument `type` can be used to choose among the 9 definitions
```
andreasnoack, line 945: We could consider calling the keyword `definition` to follow the naming of the paper. I guess `type` is what the R functions use.

src/Statistics.jl
```diff
@@ -923,31 +927,47 @@ julia> median([√1, √3, √2])
...
+in Hyndman and Fan (1996). Alternatively, `alpha` and `beta` allow reproducing
+any of the methods 4-9 defined in this paper. It is not allowed to specify both
```
andreasnoack, line 947:
```suggestion
any of the definitions 4-9 of this paper. It is not allowed to specify both
```
just to avoid also introducing "method" as another synonym on top of "definition" and "type".

andreasnoack (commented):
Looks good to me. Just some minor comments/questions

test/runtests.jl
nalimilan: Unfortunately it doesn't infer in the current state of the PR. I need to add `@inline` for it to work. But that's not great as `_quantile` has to be `@inline` (as currently) too, which means we would inline a significant amount of code.

I've pushed a commit which inlines one-line methods but doesn't inline `_quantile`, which takes a `Val` instead. This infers correctly but the code is more complex.

I also tried `Base.@constprop :aggressive`. It improves inference for most methods, but not for those taking `p::Vector` (probably due to `map`).

src/Statistics.jl
nalimilan: Good point. I've turned these into `Int`s and `Rational`s instead, and convert them to the type of `p` (which determines the return type anyway due to promotion in the calculations). In theory this could be slightly breaking for the default quantile type if you passed non-`Float64` `p` and values, but in practice I don't think it's used.

nalimilan (commented):
Woops, I completely forgot about this!

src/Statistics.jl
nalimilan: Yeah that's the R terminology. NumPy says "method" (but it takes names rather than numbers).

PR #189, opened 2025-02-07
Remove SparseArrays from deps

j-fu:
Statistics.jl  has minimum julia version 1.9, so this band-aid is not necessary anymore.

Should at once help  https://github.com/SciML/LinearSolve.jl/issues/573


nalimilan:
Thanks!

PR #197, opened 2026-01-05
fix implicit conversion of range to array

MilesCranmer:
`std` is inadvertently rerouting `std(::AbstractRange)` through `var(::AbstractArray)` rather than `var(::AbstractRange)`, since the latter does not support keywords, and the bug https://github.com/JuliaLang/julia/issues/9498 exists (patched in https://github.com/JuliaLang/julia/pull/60499, which led me to spot this downstream effect)

For example:

```julia
julia> @b 100_000_000 std(1:_)
91.392 ms

julia> @b 100_000_000 var(1:_)
5.610 ns
```

This PR fixes this so that `std(::AbstractRange)` now uses `var(::AbstractRange)`


src/Statistics.jl
```diff
@@ -214,6 +214,9 @@ function _var(iterable, corrected::Bool, mean)
     y = iterate(iterable)
     if y === nothing
         T = eltype(iterable)
+        # For truly empty iterables like `()`, the element type is `Union{}` and we
+        # intentionally throw (instead of returning NaN) for type-stability.
+        T === Union{} && throw(MethodError(var, (iterable,)))
```
nalimilan: Are you sure this is needed? AFAICT we already throw an error because `zero(::Union{})` isn't defined.

At any rate, I think it would be clearer to throw an `ArgumentError` explaining the problem instead of pretending the method doesn't exist (it does).

src/Statistics.jl
```diff
@@ -398,28 +401,42 @@ varm(iterable, m; corrected::Bool=true) = _var(iterable, corrected, m)
...
+    vv = range_varm(v, m)
+    return corrected ? vv : vv * (l - 1) / l
+end
 
 function range_varm(v::AbstractRange, m)
     f  = first(v) - m
     s  = step(v)
     l  = length(v)
-    vv = f^2 * l / (l - 1) + f * s * l + s^2 * l * (2 * l - 1) / 6
+    vv = abs2(f) * l / (l - 1) + f * s * l + abs2(s) * l * (2 * l - 1) / 6
```
nalimilan: Is this change tested somewhere?

src/Statistics.jl
```diff
@@ -398,28 +401,42 @@ varm(iterable, m; corrected::Bool=true) = _var(iterable, corrected, m)
...
-function var(v::AbstractRange)
-    s  = step(v)
-    l  = length(v)
-    vv = abs2(s) * (l + 1) * l / 12
-    if l == 0 || l == 1
-        return typeof(vv)(NaN)
+function var(v::AbstractRange; corrected::Bool=true, mean=nothing)
+    l = length(v)
+    if l <= 1
+        return _var(v, corrected, mean)
```
nalimilan: Why not keep the existing approach `return typeof(vv)(NaN)` ? It's simpler to understand.

nalimilan (commented):
Good catch!

Are you sure existing tests cover all cases in the new code? I see one line which isn't covered in Codecov and I wonder whether there might be corner cases.

src/Statistics.jl
MilesCranmer: This was required to avoid errors in nightly (was giving Unitful.jl issues); not sure how to test it in earlier Julia.

src/Statistics.jl
MilesCranmer: I think that's incorrect if `corrected == false`

MilesCranmer: Because of this:
```julia
        sum2 = abs2(value - mean::Number)
        while y !== nothing
            value, state = y
            y = iterate(iterable, state)
            count += 1
            sum2 += abs2(value - mean)
        end
        return sum2 / (count - Int(corrected))
```

src/Statistics.jl
nalimilan: Hmm then maybe this should be reported against Julia or Unitful.jl. I don't see why squaring should fail. Semantically, we shouldn't have to call `abs2` for a method which doesn't take complex numbers (since it's a range).

MilesCranmer: Ok I will revert it, but just a heads up you might need to fix it again later

src/Statistics.jl
```diff
@@ -398,28 +401,37 @@ varm(iterable, m; corrected::Bool=true) = _var(iterable, corrected, m)
...
     return vv
 end
 
-function var(v::AbstractRange)
-    s  = step(v)
-    l  = length(v)
-    vv = abs2(s) * (l + 1) * l / 12
-    if l == 0 || l == 1
-        return typeof(vv)(NaN)
+function var(v::AbstractRange; corrected::Bool=true, mean=nothing)
```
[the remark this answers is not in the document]
MilesCranmer: I just tried. I think it's actually cleaner when it's split. The combined method ends up looking like this:

```julia
function _var(A::AbstractArray, corrected::Bool, mean, dims::Dims) where Dims
    if mean === nothing
        mean = dims === (:) ? Statistics.mean(A) : Statistics.mean(A, dims=dims)
    end
    v = if dims === (:)
        varm(A, mean; corrected=corrected)
    else
        varm(A, mean; corrected=corrected, dims=dims)
    end
    return dims === (:) ? real(v) : v
end
```

unless you want the scalar `varm` to accept a `dims` keyword, but I think that might be a bit beyond the PR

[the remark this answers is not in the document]
MilesCranmer: Oh wait. Actually... I think that might be another bug. Because we currently have

```julia
varm(A::AbstractArray, m::AbstractArray; corrected::Bool=true, dims=:)
varm(A::AbstractArray, m; corrected::Bool=true)
```

Yeah... that's a bug. I'll fix that too.

MilesCranmer:
Let me know if you'd prefer this:
```julia
function _var(A::AbstractArray, corrected::Bool, mean, dims::Dims) where Dims
    if mean === nothing
        mean = dims === (:) ? Statistics.mean(A) : Statistics.mean(A, dims=dims)
    end
    v = if dims === (:)
        varm(A, mean; corrected=corrected)
    else
        varm(A, mean; corrected=corrected, dims=dims)
    end
    return dims === (:) ? real(v) : v
end
```

or keep it split as-is

nalimilan:
What I would do is this:
```julia
function _var(A::AbstractArray, corrected::Bool, mean, dims::Colon)
    if mean === nothing
        mean = Statistics.mean(A)
    end
    return varm(A, mean; corrected=corrected)
end

function _var(A::AbstractArray, corrected::Bool, mean, dims::Dims) where Dims
    if mean === nothing
        mean = Statistics.mean(A, dims=dims)
    end
    return real(varm(A, mean; corrected=corrected, dims=dims))
end
```

Even if that's a bit verbose, the advantage is that it's clear that it calls `varm`, so no need to check whether the methods give the same results (which is always hard in corner cases).

src/Statistics.jl
```diff
@@ -207,13 +207,17 @@ end
...
 
 var(iterable; corrected::Bool=true, mean=nothing) = _var(iterable, corrected, mean)
 
 function _var(iterable, corrected::Bool, mean)
     y = iterate(iterable)
     if y === nothing
         T = eltype(iterable)
+        # For truly empty iterables like `()`, the element type is `Union{}` and we
+        # intentionally throw (instead of returning NaN) for type-stability.
+        T === Union{} && throw(ArgumentError("variance of an empty iterator is undefined"))
```
nalimilan: Worded that way the message would imply that we throw all the time, but that's only the case for `Union{}`.
```suggestion
        # When element type is Union{} we don't return NaN for type stability
        T === Union{} &&
            throw(ArgumentError("cannot compute variance of an empty iterator with Union{} eltype"))
```

MilesCranmer: What other iterator would have this eltype? I feel like saying it's empty is more helpful; I didn't even know that () had that eltype, I would have assumed Any

nalimilan:
Sorry at first I thought the methods you proposed used `::AbstractRange`, not `::AbstractArray`. Then can't we just keep them as they are, and just drop `var(v::AbstractRange)`?

src/Statistics.jl
nalimilan: Could also be `Union{}[]`, a generator over such a type, or any custom iterator. In the message I proposed we say both that it's empty and give the eltype. I think this could be helpful as `map` and `broadcast` can sometimes give e.g. an empty `Vector{Float64}` and sometimes an empty Vector{Union{}}` depending on whether inference succeeded to guess the eltype or not. So you could suddenly get this error on code that worked perfectly before without any obvious reason if we don't print the eltype.

src/Statistics.jl
```diff
@@ -398,28 +402,42 @@ varm(iterable, m; corrected::Bool=true) = _var(iterable, corrected, m)
...
 function range_varm(v::AbstractRange, m)
     f  = first(v) - m
     s  = step(v)
     l  = length(v)
-    vv = f^2 * l / (l - 1) + f * s * l + s^2 * l * (2 * l - 1) / 6
     if l == 0 || l == 1
-        return typeof(vv)(NaN)
+        T = typeof((abs2(f) + abs2(s) + realXcY(f, s)) / 2)
+        return T(NaN)
     end
```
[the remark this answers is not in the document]
nalimilan: I mean, why not keep `return typeof(vv)(NaN)`? We don't care about redundant calculations in that case. Also, I suspect the compiler is able to eliminate any of these.

MilesCranmer:
Well if we fully remove the `var(v::AbstractRange)` branch then we lose the fast formula `abs2(step(v)) * (l+1)*l/12` so I'm not quite sure I follow

nalimilan:
But we can still use if `m == nothing` (either with an `if` or with a separate method `m::Nothing`). Really, `varm` is just `var(... mean=...)` designed before keyword arguments were fast, the two are exactly the same.

src/Statistics.jl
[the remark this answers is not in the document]
nalimilan: Yeah we don't care about performance in this corner case, and the compiler will probably be able to eliminate all of these.

src/Statistics.jl
```diff
@@ -398,28 +402,37 @@ varm(iterable, m; corrected::Bool=true) = _var(iterable, corrected, m)
 
 ## variances over ranges
 
-varm(v::AbstractRange, m::AbstractArray) = range_varm(v, m)
-varm(v::AbstractRange, m) = range_varm(v, m)
+function varm(v::AbstractRange, m; corrected::Bool=true)
+    l = length(v)
+    l <= 1 && return _var(v, corrected, m)
+    vv = range_varm(v, m)
```
nalimilan: What I was suggesting is this:
```suggestion
    vv = if m === nothing
        s = step(v)
        abs2(s) * (l + 1) * l / 12
     else
         range_varm(v, m)
     end
```
Then I think you can just remove `var(v::AbstractRange...`, and probably even merge the contents of  `range_varm` into this method.

src/Statistics.jl
```diff
@@ -398,28 +402,37 @@ varm(iterable, m; corrected::Bool=true) = _var(iterable, corrected, m)
...
-    vv = f^2 * l / (l - 1) + f * s * l + s^2 * l * (2 * l - 1) / 6
-    if l == 0 || l == 1
-        return typeof(vv)(NaN)
-    end
-    return vv
+    # For yᵢ = f + (i-1)s with i ∈ 1:l, we want sum(abs2(yᵢ)) / (l-1).
+    sum_t = l * (l - 1) / 2
+    sum_t2 = (l - 1) * l * (2 * l - 1) / 6
+    numer = abs2(f) * l + 2 * realXcY(f, s) * sum_t + abs2(s) * sum_t2
+    return numer / (l - 1)
```
nalimilan: Sorry but I still don't understand what you did here:
- What's the advantage of splitting the expression into several variables? You can just add a line break if needed.
- Why multiply by `l - 1` only to divide by `l - 1` in the end?

nalimilan:
Thanks! So now we get more features almost without more code!

PR #206, opened 2026-06-06
Throw a clear error for degenerate-length data vectors (fixes #121)

### LINKED ISSUE #121: `quantile!` produces an error when computing quantiles of a large range
In an earlier issue @jishnub [noticed](https://github.com/JuliaLang/julia/issues/45389#issuecomment-1170061803) that certain ranges are considered nonempty despite the fact that their length is zero:

```jl
julia> r = typemin(Int):typemax(Int)
-9223372036854775808:9223372036854775807

julia> length(r)
0

julia> isempty(r)
false
```

This behavior triggers an assertion error in `quantile!` even though quantiles for such ranges are mathematically well-defined:

```jl
julia> using Statistics

julia> quantile!([0], r, [1])
ERROR: AssertionError: n > 0
```

The assertion has the comment "this case should never happen here":

https://github.com/JuliaStats/Statistics.jl/blob/0588f2cf9e43f9f72af5802feaf0af4b652c3257/src/Statistics.jl#L994

JohnCobbler:
`quantile!` throws `ArgumentError("empty data vector")` when `length` is 0. Fixes #121.

`typemin(Int):typemax(Int)` reports `length` 0. `isempty` is false, so the old emptiness check let it through to `@assert n > 0`. The guard is `length(v) == 0`.

```julia
r = typemin(Int):typemax(Int)
quantile!([0], r, [1])        # ArgumentError
quantile(r, 0.5; sorted=true) # ArgumentError
```

`quantile(r, 0.5)` with `sorted=false` still errors in `Base.copymutable`.

### DIFF
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -1026,6 +1026,10 @@ quantile!(v::AbstractVector, p::Real; sorted::Bool=false, alpha::Real=1.0, beta:
 function _quantilesort!(v::AbstractVector, sorted::Bool, minp::Real, maxp::Real)
     isempty(v) && throw(ArgumentError("empty data vector"))
     require_one_based_indexing(v)
+    length(v) == 0 && throw(ArgumentError(
+        "data vector reports length 0 but is not empty; this can happen for an " *
+        "integer range whose length overflows (e.g. `typemin(Int):typemax(Int)`) — " *
+        "collect it to a concrete vector first"))
 
     if !sorted
         lv = length(v)
@@ -1051,7 +1055,7 @@ end
 
     n = length(v)
 
-    @assert n > 0 # this case should never happen here
+    n > 0 || throw(ArgumentError("quantile is undefined for a length-0 data vector"))
 
     m = alpha + p * (one(alpha) - alpha - beta)
     # Using fma here avoids some rounding errors when aleph is an integer
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -714,6 +714,17 @@ end
     @test quantile(skipmissing([1, missing, 2]), 0.5) === 1.5
     @test quantile([1], 0.5) === 1.0
 
+    # A range whose length overflows reports length 0 while being non-empty,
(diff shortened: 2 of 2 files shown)


src/Statistics.jl
```diff
@@ -1026,6 +1026,10 @@ quantile!(v::AbstractVector, p::Real; sorted::Bool=false, alpha::Real=1.0, beta:
 function _quantilesort!(v::AbstractVector, sorted::Bool, minp::Real, maxp::Real)
     isempty(v) && throw(ArgumentError("empty data vector"))
     require_one_based_indexing(v)
+    length(v) == 0 && throw(ArgumentError(
+        "data vector reports length 0 but is not empty; this can happen for an " *
+        "integer range whose length overflows (e.g. `typemin(Int):typemax(Int)`) — " *
+        "collect it to a concrete vector first"))
```
andreasnoack, line 1032: I believe that such a vector would take up 148 exabytes, so probably not the right recommendation.

JohnCobbler:
Agreed. 148 exabytes is not a useful suggestion. The message says the length doesn't fit in Int, and asks for a collection whose length does.

nalimilan:
Thanks! Though I have the impression that printing a specific message for this particular range is overkill. There are probably many places in Statistics, and other JuliaStats packages where we would have to do the same. Yet it seems very unlikely that people would use such a range in practice. So while being correct is of course a requirement, providing a nice error message doesn't seem super important.

How about just replacing the `isempty(v)` check with `length(v) == 0`? Or maybe even better, try to fix this in Base, along the lines of https://github.com/JuliaLang/julia/issues/45389#issuecomment-1532098487.

JohnCobbler:
Fair point. The guard is `length(v) == 0` now, and this range gets the same empty-vector error as `Int[]`. I'll leave the Base `length` change to JuliaLang/julia#45389.

nalimilan (approved):

andreasnoack (approved):

PR #210, opened 2026-08-14
Test against Julia nightly in CI

### LINKED ISSUE #208: `reducedim1` removed from Base
this line now fails
https://github.com/JuliaStats/Statistics.jl/blob/e14c8f8ec37f33fc4413a25915cb443899748ae3/src/Statistics.jl#L288

after
https://github.com/JuliaLang/julia/pull/62554

andreasnoack:
Adds `nightly` to the CI version matrix so the package is also tested against nightly builds of Julia on all three platforms.

The nightly jobs initially failed because the internal `Base.reducedim1` changed signature from `reducedim1(R, A)` to `reducedim1(R)` on nightly (#208). Since the second argument was always unused and the function is a one-liner, this PR defines it locally in Statistics.jl, which works on all supported Julia versions.

Fixes #208

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,7 @@ jobs:
           - '1' # latest v1 version of Julia
           - 'lts' # latest LTS build of Julia
           - 'pre' # latest prerelease build (RCs, betas, and alphas) of Julia
+          - 'nightly' # nightly build of Julia
         os:
           - ubuntu-latest
           - macOS-latest
--- a/src/Statistics.jl
+++ b/src/Statistics.jl
@@ -257,10 +257,6 @@ function _var(iterable, corrected::Bool, mean)
     end
 end
 
-# Whether R has a single element along its first dimension. Defined here instead
-# of using the internal Base.reducedim1 whose signature changed in Julia 1.14
-reducedim1(R) = length(Base.axes1(R)) == 1
-
 centralizedabs2fun(m) = x -> abs2.(x - m)
 centralize_sumabs2(A::AbstractArray, m) =
     mapreduce(centralizedabs2fun(m), +, A)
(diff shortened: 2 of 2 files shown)


andreasnoack:
Given that Julia CI tests against this package, we should probably test on nightly.

devmotion (commented):
I'm generally not convinced that packages should test with nightly (or even pre) at all (see e.g. https://discourse.julialang.org/t/why-do-packages-run-continuous-integration-tests-on-julia-nightly/101208/3), but it's arguably a bit different here since Statistics is a stdlib. So I'm fine with adding it to CI.

Regarding `reducedim1`: it seems it's only used in a single place, so why defining a function for it at all? Couldn't we just inline the check?

devmotion (approved):

andreasnoack:
> Regarding reducedim1: it seems it's only used in a single place

Indeed. I had just assumed that it was used more than once. I've removed the helper function.

KristofferC:
Can we get a new release with this?

andreasnoack:
I believe that https://github.com/JuliaLang/julia/pull/62770 included this, but I realize that I hadn't bumped any versions.

KristofferC:
This stdlib is "upgradable", pkg will resolve it from the registry so it needs versioning like a normal package. The one bundled with julia is just a fall back. It is even bundled at all because triage decided that removing any stdlib ever from the julia repo would be breaking. Anyway, I made a new release.

andreasnoack:
I realize that I have no idea how this works for stdlibs. Where did you make that release? I don't see it in General and I don't see any new tags or releases here.

KristofferC:
https://github.com/JuliaStats/Statistics.jl/commit/1bb9f77dcba5b0c8ba1a0ea28680eaabf18b6e85, https://github.com/JuliaRegistries/General/pull/166296 

But it didn't get auto merged it seems because I accidentally skipped a version when bumping the version number.

KristofferC:
(only DelimitedFiles and Statistics are currently upgradable independently of the Julia version)

PR #213, opened 2026-09-02
test: Do not assert which NaN `mean` returns

maleadt:
Julia does not specify the sign or payload of a NaN produced by floating-point arithmetic, and LLVM explicitly permits both to vary. RISC-V canonicalizes NaN results from ordinary arithmetic, so the exact-equality assertions fail there even though `mean` correctly returns a NaN (JuliaLang/julia#56672).

Check only that the negative-NaN inputs produce a NaN, consistent with the existing positive-NaN coverage.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 keywords = ["statistics"]
 license = "MIT"
 desc = "Basic statistics for Julia."
-version = "1.11.4"
+version = "1.11.5"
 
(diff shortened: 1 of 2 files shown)


maleadt:
Included a version bump because this matters for Julia Base CI on RISC-V.

nalimilan (approved):

maleadt:
I don't have permissions on here, so I'd appreciate if somebody could merge and tag.

nalimilan:
Done: https://github.com/JuliaRegistries/General/pull/166901

